PR #1, opened 2018-11-14
An initial idea

wooorm:
Hi folks! I took some time today to draft out an early early (really early) idea. It doesn’t really work, but it gives us a starting point. You can clone it and run `node example`, and try changing the given heading to something else, to see how it works.

This is mainly inspired by how HTML is parsed. How it’s specced, but also how parses5 implements that spec. HTML, even though it’s pretty structured, has the same problem as markdown: everything is valid. That’s why I thought we’d start with this.

With this concept, I tried to get something of an AST done. Some initial positional info. One initial node (ATX heading), and some ideas on what’s needed for the block and inline states.

This is not something that should be merged, it‘s the first thing we throw away to build something better!


wooorm:
P.S. for everyone not on the review list reading this: please review the code and share initial ideas! I’d love to develop this transparently and in the open, we could use fresh ideas on this!

zeusdeux:
I'll pop in on this next week. Currently dealing the pre-holiday work pile up.

lib/characters.js
```diff
@@ -0,0 +1,113 @@
...
+exports.uppercaseX = 88 //  'X'
+exports.uppercaseY = 89 //  'Y'
+exports.uppercaseZ = 90 //  'Z'
+exports.leftSquareBracket = 91 //  '['
+exports.backslash = 92 //  '\'
+exports.rightSquareBracket = 93 //  ']'
+exports.caret = 94 //  '^'
+exports.underscore = 95 //  '_'
+exports.graveAccent = 96 //  '`'
+exports.lowercaseA = 97 //  'a'
```
ChristianMurphy: These (export lowercase) appear to be defined but not used.
Is there an expectation that downstream developers will use this?

wooorm: I found this one in babel. They actually have a plugin to inline them all, but before doing that I just thought I’d include all these characters at first :p

wooorm:
@ChristianMurphy Syntax extensions should go to micromark, but they could have a more gnarly interface. And I think there should be nicer plugins on a remark level that then add the plugins to the used micromark parser

lib/characters.js
```diff
@@ -0,0 +1,113 @@
+exports.eof = null
```
fazouane-marouane: What would be the need to have this file? Will all texts have only alphanumeric characters?
I also saw usage of `fromCharCode` and `charCodeAt`. Wouldn't it be preferable to acknowledge other character sets? (I'm thinking `fromCodePoint` and `codePointAt`)

example.js
```diff
@@ -0,0 +1,7 @@
+var Micromark = require('.')
+
+var m = new Micromark()
+
+var res = m.write('# Hello!')
+
+console.log('res: ', res)
```
fazouane-marouane: It seems to me that `res` will always be "undefined"

example.js
```diff
@@ -0,0 +1,7 @@
+var Micromark = require('.')
+
+var m = new Micromark()
+
+var res = m.write('# Hello!')
```
fazouane-marouane: I don't know why but we can't write more than once and continue emitting tokens or maybe that I didn't test it well enough.
Will those two codes do the same thing: `m.write('# '); m.write('Hello!');` and `m.write('# Hello!');` ?

lib/atx-heading.js
```diff
@@ -0,0 +1,315 @@
...
+    self.consume()
+  } else {
+    info.closingSequenceBefore = null
+    info.closingSequence = null
+    info.closingSequenceAfter = null
+    self.reconsume(CONTENT_STATE)
+  }
+}
+
+function bogusState() {
```
fazouane-marouane: Is this the equivalent thing for backtracking?

lib/characters.js
wooorm: This file exists to have the codes used by markdown syntax. I copied it from babel, where they use a babel transform to compile them away and use the codes directly, as markdown only uses ASCII markers, we’re fine using charCodeAt I think!

example.js
wooorm: Yes, it is, this was just for debugging purposes 😉

example.js
wooorm: Ahh, right. I didn’t check that yet. But correct, I hope to support streaming, so that you could do multiple writes and those examples are indeed equivalent! 👍

lib/atx-heading.js
wooorm: Yes!

ChristianMurphy:
> there should be nicer plugins on a remark level that then add the plugins to the used micromark parser

So the flow for something like https://github.com/djm/remark-shortcodes would be to add a micromark lexing plugin, then use the micromark output to feed a custom remark parser?

wooorm:
@ChristianMurphy Yep!

PR #2, opened 2018-11-20
Add typescript support to the repository

zeusdeux:
## This PR adds the following

- `.nvmrc` so that we all use the same version of node (11.2.x)
- the following dev tooling
  - `tsc` with a strict tsconfig setup
  - `tslint` for linting
  - `prettier` for code formatting
  - ~`mocha` as our test runner with TS support setup~
  - ~`sinon` for spies, stubs, etc~
  - ~`chai` for assertions~
  - `mocha`, `sinon` and `chai` have been superseded by `jest`
- a dummy `lib/index.ts`
  - named the folder `lib` for consistent with other unifiedjs repos
  - this is where I expect our code will sit
- a dummy unit test in `lib/__tests__/`
- all the tooling wired up

## Good to haves (not added yet)

- lint-staged to prettify and lint when staging a file
- husky to run lint-stage on a git hook like pre-commit

## Open questions

- which test runner? (currently: `mocha`)
  - switched to `jest` ✅ 
- which assertion library? (currently: `chai`)
  - `jest` all the things! ✅ 
- what dev tooling? (currently: `tsc`, `tslint` and `prettier`)
  - current tooling is ✅ 
- what directory structure? (currently: `lib` will hold code, `test`
will hold tests)
  - `lib` for code and `__tests__` folder next to the code that holds the tests ✅ 
- what linting rules to use? (currently: tslint-recommended + my
overrides)
- What license? (currently: ISC)
  - `MIT` ✅ 
- What version? (currently: `0.0.1`)
- Author and contributor data in `package.json`
- Do we start by defining interfaces, potential extension points, etc? The design ie.
- Do we start with prototyping directly?
  - If so how do we merge #1 into this if we choose to do so?


package.json
```diff
@@ -0,0 +1,48 @@
...
+  },
+  "keywords": [
+    "markdown",
+    "lexer",
+    "lexing",
+    "unifiedjs"
+  ],
+  "author": "",
+  "contributors": [],
+  "license": "ISC",
```
zcei: I think the rest of the unified projects are MIT. That would be my preference at least for a healthy ecosystem

package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "micromark",
+  "version": "0.0.1",
```
zcei: We have to start at `2.0.0` as we got the `micromark` package from someone who has already published a `1.0.0`.

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+  "compilerOptions": {
+    "target": "ES2017",
```
ChristianMurphy: I think producing an ES2017 or even ES2018 version is a good idea. :+1:
Could we also build an ES5 version?

See: https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages

package.json
wooorm: If we’d use the scope, `@micromark`, we can start at 1.0.0. If we don’t scope, we need to start above 2.0.0. I’d say leave it at this and we can decide later.

vhf:
> - which test runner? (currently: `mocha`)
> - which assertion library? (currently: `chai`)

I'd suggest Jest. This way no need for one tool to run tests (mocha), one for the assertion (chain), one for the coverage, etc. It also has a very useful CLI to run/watch partial tests, and it runs faster than mocha.

👍 / 👎 ?

zeusdeux:
@vhf done! ^_^
@ChristianMurphy I haven't addressed your comment yet as I am yet to read the article. Will do so once I reach my destination ^_^

zeusdeux:
@zcei @wooorm @vhf @ChristianMurphy Is this good to go?

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "printWidth": 100,
+  "semi": false,
+  "singleQuote": true,
+  "trailingComma": "none"
+}
```
fazouane-marouane: Are these personal preferences or are they inherited from related projects?

zeusdeux: @fazouane-marouane Inherited from other projects under the unifiedjs umbrella. 
I can update this if you notice any drift already. If not, we can merge it and modify it in the future.

wooorm: Let’s try this! We can change it later if need be.

.nvmrc
```diff
@@ -0,0 +1 @@
+11.2
```
wooorm: Is this handy? Isn’t it nice that people are on different Node versions, means more configurations used, so less bugs?!

package-lock.json
```diff
@@ -0,0 +1,5675 @@
+{
```
wooorm: I personally like not having locks in libraries. But I’m up for changing my mind.

tsconfig.json
wooorm: @zeusdeux I’m also interested in this one.

package-lock.json
fazouane-marouane: It’s particularly interesting to have them if everything is bundled (so practicly no dependencies, only devdependencies). But bundling seems weird for a nodejs library and lockfiles will not be honored by npm while installing this lib within another project. I agree that the lock maybe omitted, unless we opt not to have dependencies but only devdependencies.

.nvmrc
zeusdeux: I did this only so that `npm install`s don't lead to an updated `package-lock.json` due to variance in `npm` version. But given the conversation below around the lock file, I am down to remove this if we remove the lock file as well. It's predicated on what we decide below.

package-lock.json
zeusdeux: @wooorm What's your rationale behind not including a lock file for libs?

Rokt33r: I think it just makes commit history dirty. This comment would be helpful.
https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

---

And, we could omit it easily by adding `.npmrc`

```
package-lock=false
```

wooorm: Yeah, I think Sindre sums it up nicely. And I don’t agree with the Yarn post by James.

I ignore `package-lock.json` through an [`.npmrc`](https://github.com/remarkjs/remark-rehype/blob/master/.npmrc), and `yarn.lock` from [`.gitignore`](https://github.com/remarkjs/remark-rehype/blob/07469598e3a2c794f51e0741b8213066b3c702c9/.gitignore#L8).

zeusdeux: My counterargument would be as follows — we can’t reliably know what versions of the dependencies our user will install unless we have exact versions in our `package.json`. The possibility of the code in this package running with different versions of its dependencies is inevitable. Therefore, irrespective of the existence of a lockfile, bugs caused by this are inevitable. Also, @sindresorhus’ comment about publishing a buggy version due to variance in dependency versions doesn’t go away by removing the lock file. On the contrary, having a lockfile helps figure out faster if a version drift caused the bug. 

Having a lockfile at least makes the dependency tree known during development. This comes with benefits such as faster installs, `npm ci`, potentially faster debugging as we can rule out dependency update related bugs faster, etc. 

As I see it, having a lockfile doesn’t impact our users but gives us certain benefits during dev. Whereas not having it rids us of these benefits while not having any impact on our users. Therefore, I’d vote to commit the file.

tsconfig.json
zeusdeux: @ChristianMurphy so if I understand correctly, you’d want `main` to be the entry for the code transpiled to ES5 + CJS and `module` to be code transpiled to ES2017/ES2018+ES6 modules?

package-lock.json
fazouane-marouane: The argument was about not having the same view as your user’s. Having a lockfile is excellent for developers but gives false confidence once your ci installs the exact same dependencies as in every dev’s machine. The ci should reflect the real world, what would happen if somebody installs your package and tries it. If one of your dependencies introduces a breaking change, devs and ci won’t see it but your users will.

You can always use a lockfile without committing it and you’ll get all benefits of a lockfile locally. As for having exact dependency versions, it won’t solve the issue: one of you dependencies may not use exact version for one of its dependencies.
The only way to guarantee that users will always get the same package without conflict with the external world would to either bundle the lib or to not use dependencies. Not having the lockfile inside the repo will make us detect issues as they happen.

@wooorm @zeusdeux @Rokt33r would having at least one frequent ci job/sanity check ignore the lockfile be a good compromise ?

zeusdeux: I’ll withdraw my vote in support of committing the lockfile as there’s momentum against it and it feels like I’m fueling bikeshedding. 
I’ll remove em in a commit that follows. If we find we see value in having it in the future, we can add it back in. 
Sound good?

tsconfig.json
ChristianMurphy: > main to be the entry for the code transpiled to ES5 + CJS 

:+1: exactly

> `module` to be code transpiled to ES2017/ES2018+ES6 modules?

Close, from the babel article:

> However, the sole intention of this field is ESM, not anything else. The Rollup docs specify that the `module` field makes it clear that it's not intended for future JavaScript syntax.
>
> Despite this warning, package authors invariably conflate the use of ES modules with the JavaScript language level they authored it in.

https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages#conflating-javascript-modules-and-es2015

publishing ES2015+ under `module` can break rollup builds and create react app v1 builds.

One of the points of the articles is there is not _yet_ a standard way to ship ES2015+ in packages.
The proposed syntax right now `main-es` or `es`.
Which runs along side the existing `main` and `module` (both in ES5).

wooorm: Tom just wrote a nice post about how to bundle this all, except the typescript part: https://macwright.org/2018/11/26/small-modules.html, it mentions `microbundle` as well, just like the babel post above.

.nvmrc
zeusdeux: Update: this is now nuked. ^_^

tsconfig.json
ChristianMurphy: Looks like microbundle supports Typescript out of the box now.
https://github.com/developit/microbundle/pull/16

package-lock.json
wooorm: Sweet, we can revisit it later if need be!

tsconfig.json
wooorm: Nice, I’m up for trying it, what do y’all think?

wooorm: Or maybe the current setup is good enough already?

zeusdeux: @ChristianMurphy haha I was going to bring up the same point from the blog if you'd have said yes wrt `module`. Since there isn't consensus on that now, I'll do the following:

1. `main` will point to an entry for code that is transpiled to ES5 and uses CommonJS module system
2. `module` will point to an entry for code that is transpiled to ES5 but uses ES6 module system so that bundlers can do cool things like DCE/tree-shaking
3. `main-es` and `es` will both point to an entry for code that is transpiled to ES2017 (I don't see ES2018 in the compiler options and ESNext is too aggressive imho) and uses ES6 module system

How does that sound @ChristianMurphy @wooorm?
Also, any opinions on including a umd build as well?

wooorm: I like that. And :-1: to umd, I don’t see a lot of value for that in such a low-level project.

package.json
```diff
@@ -0,0 +1,64 @@
...
+    "type": "git",
+    "url": "git+https://github.com/micromark/micromark.git"
+  },
+  "keywords": [
+    "markdown",
+    "lexer",
+    "lexing",
+    "unifiedjs"
+  ],
+  "author": "",
```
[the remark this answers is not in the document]
zeusdeux: done.

package.json
```diff
@@ -0,0 +1,64 @@
...
+  "types": "built/lib/index.d.ts",
+  "module": "lib/index.ts",
+  "scripts": {
+    "build": "rimraf built && tsc",
+    "watch": "tsc --watch",
+    "lint": "tslint -p ./tsconfig.json",
+    "test": "jest",
+    "prettier": "prettier --config ./.prettierrc --write \"{bin,lib}/**/*.{ts,json}\""
+  },
+  "repository": {
```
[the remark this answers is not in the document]
zeusdeux: Done.

package.json
```diff
@@ -0,0 +1,64 @@
...
+  "keywords": [
+    "markdown",
+    "lexer",
+    "lexing",
+    "unifiedjs"
+  ],
+  "author": "",
+  "contributors": [],
+  "license": "MIT",
+  "bugs": {
```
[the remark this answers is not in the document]
zeusdeux: Done.

package.json
```diff
@@ -0,0 +1,64 @@
...
+    "url": "https://github.com/micromark/micromark/issues"
+  },
+  "homepage": "https://github.com/micromark/micromark#readme",
```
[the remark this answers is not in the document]
zeusdeux: 👍

tsconfig.json
zeusdeux: Instead of adding yet another tool, I have chosen to use the typescript compiler to do this as it supports all of this by default. A commit shall follow in the next few minutes.

zeusdeux:
Question: Since we are compiling down to ES5 as our `main`, do we want to take care of polyfilling missing methods ourselves as well or leave that to the consumers of this package?

/cc @wooorm @vhf @zcei @ChristianMurphy @fazouane-marouane @Rokt33r

tsconfig.json
zeusdeux: The commit is blocked by https://github.com/micromark/micromark/pull/2#issuecomment-441693780 :(

package.json
```diff
@@ -0,0 +1,65 @@
...
+    "lexer",
+    "lexing",
+    "unifiedjs"
+  ],
+  "author": "Titus Wormer <tituswormer@gmail.com> (https://wooorm.com)",
+  "contributors": [
+    "Titus Wormer <tituswormer@gmail.com> (https://wooorm.com)",
+    "John Otander <johnotander@gmail.com> (http://johnotander.com)",
+    "Stephan Schneider <stephanschndr@gmail.com>",
+    "Victor Velder <victor@draft.li> (https://draft.li)",
```
vhf:
```suggestion
    "Victor Felder <victor@draft.li> (https://draft.li)",
```

ChristianMurphy:
> do we want to take care of polyfilling missing methods ourselves as well or leave that to the consumers of this package?

I'd leave polyfilling up to down stream consumers.
I've run into issues where babel-polyfill will conflict with other polyfills applied by downstream consumers, causing warning and errors out of the blue. (https://github.com/uPortal-contrib/uPortal-web-components/pull/36)
The one thing to watch out for is `regenerator-runtime`, needed for `async await`, is assumed to exist by the babel transform, and needs to be added globally or manually imported, if it is not polyfilled by `babel-polyfill` (https://github.com/uPortal-contrib/uPortal-web-components/issues/43)

zeusdeux:
Could you take a look at a466320 folks? 🙇

zeusdeux:
The typescript compiler does what it needs to support its own syntax in the configured `target` environment. Which, as @fazouane-marouane pointed out, includes polyfilling _some_ things.

zeusdeux:
@wooorm yupp. Since we don't polyfill anything, even for `es5-cjs` only node 10 or above are support _out of the box_. This obviously is predicated on how many ES2018 features we _do_ use. For e.g., if we use `Promise.finally`, it'll only work on node 10+ out of the box. Nevertheless, given that the probability of using that and the new regex features is low, I am down to remove the engines key.

Also, I am assuming you're talking about the engines key haha

zeusdeux:
Setting up tooling is fun eh? 😬

wooorm:
Incredibly fun 😉
And yes, I’m talking about `engines`! I like dropping it, because it seems that there’s always more/less versions actually supported. But maybe this is true!

zeusdeux:
and done.

zeusdeux:
Any final comments anyone? If not, I'll merge this tomorrow.

wooorm:
Yay!

PR #3, opened 2018-11-29
Ignore lockfiles

vhf:
Since [we decided](https://github.com/micromark/micromark/pull/2#discussion_r236091486) not to commit a lockfile, I suggest adding npm and yarn lockfiles to `.gitignore`. That way we don't have to manually ignore them while commiting or accidentally commiting them, and developers are free of using the package manager of their choice.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,10 @@ node_modules/
 .env
 .envrc
 
+# lockfiles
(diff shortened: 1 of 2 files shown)


vhf:
@ChristianMurphy Good call!

PR #4, opened 2018-12-05
Migrate #1 to use typescript

fazouane-marouane:
Following PR #2, and to allow the discussion around the initial idea (PR #1 ) to continue, I decided to migrate code on `trying-this-out` branch to typescript. I didn't know if something was in the works or not.
Anyway, no changes to logic have been done and the example code has been transformed into a unit test + a snapshot data.
Please, tell me what do you think of this.

Thanks,
Marouane


fazouane-marouane:
@wooorm wouldn't it be better to merge the changes on the original PR #1 instead of master ?

wooorm:
cc @micromark/members

wooorm:
> wouldn't it be better to merge the changes on the original PR #1 instead of master ?
> [@fazouane-marouane](https://github.com/micromark/micromark/pull/4#issuecomment-444437857)

Maybe? The typescript base is already in master so I think it makes more sense to base it off of that (now the diff view also doesn’t show unrelated changes that are already reviewed)!

wooorm:
Yes, 👍 I’d like some ideas from TS aficionados (not me 😬)

lib/index.ts
```diff
@@ -1,28 +1,99 @@
...
+  public context: ContextType = 'block'
+  public stateHandlers: any
+  public contextInfo = {}
+  public state: string = ''
 
-  return inputString
-    .padStart(strLen + Math.floor(maxLength / 2), fillWith)
-    .padEnd(strLen + maxLength, fillWith)
-}
+  public block: any = block
```
zeusdeux: I'd recommend exporting `interface`s that show the shape of `block`, `atxHeading` and `paragraph` from the respective files and using that as the type here instead of `any`

lib/atx-heading.ts
```diff
@@ -0,0 +1,307 @@
+import * as c from './characters'
+
+const maxOpeningSequenceBeforeSize = 3
+const maxOpeningSequenceSize = 6
+
+const T_SPACE = 'space'
+
+export enum StateType {
```
zeusdeux: How do you see this export being consumed?

lib/atx-heading.ts
```diff
@@ -0,0 +1,307 @@
...
+  info.openingSequenceAfter = null
+  info.content = null
+  info.closingSequenceBefore = null
+  info.closingSequence = null
+  info.closingSequenceAfter = null
+
+  tokenizer.reconsume(StateType.OPENING_SEQUENCE_BEFORE_STATE)
+}
+
+function openingSequenceBeforeState(tokenizer: any, code: number) {
```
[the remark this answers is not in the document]
zeusdeux: I think `tokenizer` is `tokenizer: Tokenizer`

lib/index.ts
```diff
@@ -1,28 +1,99 @@
...
+
+    // TODO handling code === null
+    // tslint:disable-next-line:no-console
+    console.log('consume: %s', this.state, this.now(), [String.fromCharCode(code!)])
+
+    this.column++
+    this.offset++
+  }
+
+  public reconsume(state: string) {
```
zeusdeux: `state` can be a `keyof typeof ATXStateTypes | BlockStateTypes | ParagraphStateTypes` where each of these `StateTypes` is the exported `enum` from each of the respective files.

fazouane-marouane:
Thanks for the reviews 😊 I solved some/most of them. The structure seems to be a bit clearer now, and some parts still need some enhancements:
- [ ] contextInfo for one (more typing, maybe no or less mutations)
- [ ] type states
- [ ] tweak usage of reconsume/consume/next/bogus just enough to avoid having deep nested calls
- [ ] and also try to avoid circular dependency between Tokenizer and the state handlers

Tomorrow, I'll try fixing these points and the comments that I forgot to handle.

wooorm:
@fazouane-marouane anything we can do to help move this along?

fazouane-marouane:
@wooorm I only had time this evening to work on it. I solved the deep nesting of calls and a part of decorrelating Tokenizer from the context handlers. There are some small stuff to rework and it should be good.

For now, since I changed a lot of stuff, can anyone please take another look and give some feedback?

Also, I think that some parts of the code can benefit from Generator Functions (`function *()`). Should I refrain from using them or are people here familiar with it ?

wooorm:
I’m :+1: to generators, why not?! Think it would make sense.
Also: this is really wonderful already!

fazouane-marouane:
Thanks for your encouragement 😊
I added some more typings (mainly the ContextInfo part) and added support to multiple `write()`+ `end()`.

We now have generators. Which means that we need to add [tslib](https://github.com/Microsoft/tslib) as a dependency since without it typescript duplicates its polyfills on each file that need them. However, it generates `import * as tslib_1 from "tslib";` which is not that treeshakable 😢 

That been said, 1/ Typing states and 2/ removing the residual usages of `tokenizer` inside the contextHandlers should normally be the last steps.

wooorm:
Nice work!

> We now have generators. Which means that we need to add tslib as a dependency since without it typescript duplicates its polyfills on each file that need them. However, it generates import * as tslib_1 from "tslib"; which is not that treeshakable 😢

I would like to mention that one of the goals for micromark is to be small when bundled, from the readme:

> small in file size, max 10 kB minzipped, and tiny in memory use

So if including generators is making it very bulky, I may have reservations. Any idea on the costs?

fazouane-marouane:
I updated the code to use constants + union types instead of enums since they consumed a lot of size on the final bundle. I also found a workaround for tslib (See my comment in https://github.com/Microsoft/TypeScript/issues/16999 ).

So regarding bundle size, right now, when bundling the es5+esm version using webpack4 (for example) we get **9.92KiB** minified (**2.81KiB** min+gziped). Without generators, it's **6.17KiB** min/**2.00KiB** mingzip).

image


Bundling of the esm+es2018 version gives **6.19KiB** minified (**2.04KiB** min+gziped).  Without generators, it's **5.97KiB** min/**1.99KiB** mingzip)

image

PS: This is when we use `removeComments` to remove Microsoft's lengthy Apache 2 license comment. Otherwise, it's **3.25KiB** for es5 and **2.89KiB** for es2018 (both min-gziped).

## tl;dr
All in all, the effective cost of generators is around **+1.15KiB** for es5 and **+0.90KiB** for es2018 (both minified+gzip). @wooorm  Is this acceptable enough?

Also, should we keep track of the size budget and do you have some suggestions regarding this ?

wooorm:
That seems acceptable! We’ll have to pull in some more weight for things like entities and the like tho, so I think we should make sure to revisit this later.

For size checking, [`size-limit`](https://github.com/ai/size-limit) may work!

Murderlon:
I recommend https://github.com/siddharthkp/bundlesize to keep track of bundle size :)

68747470733a2f2f63646e2e7261776769742e636f6d2f7369646468617274686b702f62756e646c6573697a652f6d61737465722f6172742f7374617475732e706e67

Murderlon:
btw @fazouane-marouane awesome work on this so far 👍

fazouane-marouane:
Both tools seem good! I bet they'll be of help. I'll take a deeper look at how they are used later, I'm glad that we don't need to write manual scripts to be able to track bundle size.

As for the current PR, do you have some change requests or some recommendations? I still have some free time this week to spend on it. Next week there will be less screen time for me because of the holidays.

@wooorm Since the current state of the parser is 100% based on your original PR, do you think that we should explore a bit more this PR before going to master? If this is the case, what's the next phase ?

Murderlon:
@micromark/members looks like it's ready to be merged 🎉

authors @wooorm / @fazouane-marouane, anything else? If not I think you can do the honors.

wooorm:
@Murderlon Yep, on my todo to merge for today or tomorrow!

wooorm:
Awesome, it landed! 🎉🎉🎉

Thanks everyone for working on this, and specifically thanks to @fazouane-marouane. Great work! 💪

PR #10, opened 2019-09-20
Add support for fenced code blocks

Murderlon:
My first time experimenting with parsers and TypeScript.

- [x] Valid and invalid code blocks.
- [ ] Info string.


wooorm:
Still wip?

PR #16, opened 2020-09-10
[WIP] migrate to typescript

ChristianMurphy:
Micromark has a number of complex/opaque types used in it's internals.
This attempts to document and validate these types using TypeScript.

In particular some complex and currently undocumented types include `Events`, `Token`, `Effects`, `ok`, and `nok`.

In addition, micromark's current usage pattern includes directly accessing internal files, meaning that for TypeScript users, most (if not all) files will need types.


lib/types.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface NotOkay {
+  (code: number): void
+}
+
+export interface Parser {
+  hooks: {
+    [key: string]: unknown
+  }
+  flow: (something: unknown) => unknown
+}
```
ChristianMurphy, line 42: These are currently reconstructions based off code flow.
Guidance on one the specifics and meaning of these types are welcome.

lib/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  lazy: unknown
+  containerState: {
+    marker: number
+    type: Type
+    initialBlankLine: unknown
+    size: number
+    _closeFlow: unknown
+    furtherBlankLines: unknown
+  }
+}
```
ChristianMurphy, line 84: These are currently reconstructions based off code flow.
Guidance on one the specifics and meaning of these types are welcome.

wooorm:
Awesome! I see this effort as useful, as the current code has only seen my eyes and is young, so there’s much that can be better!
Working on types will undoubtedly catch a bunch of things I missed!

Regarding this project as TS, well, we talked about our opinions on TS a lot already, I’m not really for it, in short:

I feel strongly that the bundle size of mm should be as small as possible. The version of mm that lived here before was written in TS and while it only supported a couple of markdown constructs, it was already over the size that mm is now. This may indicate that a project in TS is bigger than a project by hand in JS. (Practically, for ESM the [scripts](https://github.com/micromark/micromark/tree/main/script) would also need to be fixed)

I also see that less folks understand TS compared to JS (including myself). Especially important for folks who want to build extensions and look at this project for examples.

Any reason to not do the jsdoc approach?

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  _break?: boolean
+  _tokenizer?: unknown
+  _events?: Event[]
+  _subevents?: Event[]
+  _contentTokenized?: unknown
+  _close?: unknown
+  _marker?: unknown
+  _open?: unknown
+  _side?: number
+  _tabs?: unknown
```
wooorm, line 28: Generally, tabs and spaces behave the same, but in the case of a hard break through trailing spaces (`  \n`), tabs don’t work.

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export type Type = ValueOf<typeof typesObject>
+
+export type Event = [name: string, token: Token, context: unknown]
+
+export interface Effects {
+  enter: (type: Type) => Token
+  consume: (code: number) => Token
+  exit: (type: Type) => Token
```
wooorm, line 38: Enter and exit define where tokens start and end. Consume deals with a character, and moves to the next.

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  attempt: (
+    one: unknown,
+    two: unknown,
+    three?: unknown
+  ) => (code: number) => unknown
```
wooorm, line 43: Attempt deals with several values, and tries to parse according to those values. If a value resulted in `ok`, it worked, the tokens that were made are used, and `returnState` (two) is switched to. If the result is `nok`, the attempt failed, so we revert to the original state, and `bogusState` (three) is used.

The values that can be given are:
- An object of “hooks”, where the keys are character codes and the values are either a single construct or an array of constructs. If the first character that is attempted matches, the corresponding constructs are attempted (see next), otherwise: bogusState
- A list of constructs, which will be attempted in order. If one results in `nok`, we revert and the next one is attempted
- A single constructs (an object with a `tokenize` function, and potentially some other stuff)

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+    one: unknown,
+    two: unknown,
+    three: unknown
+  ) => (code: number) => unknown
+  check: (
+    one: unknown,
+    two: unknown,
+    three?: unknown
+  ) => (code: number) => unknown
+  lazy: (one: unknown, ok: Okay, nok: NotOkay) => void
```
wooorm, line 54: check, interrupt, and lazy are all “checks”: they are like attempt, but when succesful they *also* revert.
This is useful to see what lies ahead.

[interrupt](https://github.com/micromark/micromark/search?q=interrupt&unscoped_q=interrupt) is used for stuff right after a line of content.
[lazy](https://github.com/micromark/micromark/search?q=lazy&unscoped_q=lazy) is used for lines that were not properly preceded by the container.

interrupt and lazy work by setting a property on the parser (both their name). Tokenizers can then exit early if that is detected (e.g., when fenced code is started, it’s fenced code: no need to continue parsing more), or in cases where constructs cannot exist (setext heading underlines can’t be lazy)

The container parser (document) also sets these properties in certain cases directly.

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export interface Okay {
+  (code: number): void
+}
+
+export interface NotOkay {
+  (code: number): void
+}
+
```
wooorm, line 64: Okay and NotOkay are generally like other states-as-functions: they receive a code and return a function.

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface Parser {
+  hooks: {
+    [key: string]: unknown
```
wooorm, line 67: values are one construct or an array of constructs

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+  flow: (something: unknown) => unknown
```
wooorm, line 69: Some more: https://github.com/micromark/micromark/blob/307e0a59506cebbfe7eaf3d75849c3e8d94f937f/lib/parse.js#L98

These are the available tokenizers: https://github.com/micromark/micromark/blob/307e0a59506cebbfe7eaf3d75849c3e8d94f937f/lib/util/create-tokenizer.js#L13

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  defined: unknown[]
```
wooorm, line 70: This containers unique normalized identifiers for definitions:

https://github.com/micromark/micromark/blob/307e0a59506cebbfe7eaf3d75849c3e8d94f937f/lib/tokenize/definition.js#L28

It does some rather complex things to how references are parsed 😓 https://github.com/micromark/micromark/blob/307e0a59506cebbfe7eaf3d75849c3e8d94f937f/lib/tokenize/label-end.js#L192

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  lazy: unknown
+  containerState: {
+    marker: number
+    type: Type
+    initialBlankLine: unknown
+    size: number
+    _closeFlow: unknown
+    furtherBlankLines: unknown
+  }
+}
```
wooorm, line 85: Also this stuff? https://github.com/micromark/micromark/blob/307e0a59506cebbfe7eaf3d75849c3e8d94f937f/lib/util/create-tokenizer.js#L36

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  events: Event[]
+  interrupt: unknown
+  lazy: unknown
+  containerState: {
+    marker: number
+    type: Type
+    initialBlankLine: unknown
+    size: number
+    _closeFlow: unknown
+    furtherBlankLines: unknown
```
wooorm, line 83: containerState is used in the document (container) tokenizer, because generally we can parse in one go (e.g., an atx heading), but for lists and block quotes we parse a part of it (e.g., `> `), then we do the rest of the line, and then at the next line look for another marker. As we parse separate “runs” of content, information needs to be stored somewhere, and I came up with this.

All properties are used by lists

_closeFlow is a way to communicate from lists to the tokenizer that the flow is closed, but we do continue the list. That’s useful when finding a new list item, because the previous one needs to be closed, but the list remains open.

ChristianMurphy:
> The version of mm that lived here before was written in TS and while it only supported a couple of markdown constructs, it was already over the size that mm is now

The difference in size doesn't necessarily have to do with TS.
You re-architected the parser and compiler.

> I also see that less folks understand TS compared to JS (including myself). Especially important for folks who want to build extensions and look at this project for examples.

TypeScript is JavaScript, with annotations.
It's true that adding annotations technically makes it a superset/new language, it's not nearly as much of a learning curve as you make it out to be.
Take https://github.com/micromark/micromark/pull/16/commits/d286abe496363d0bddb097a52f0f553a63c978c0 for example:

```js
function tokenizeDefinition(
  effects,
  ok,
  nok
)
```

becomes

```ts
function tokenizeDefinition(
  effects: Effects,
  ok: Okay,
  nok: NotOkay
)
```

and

```js
function start(code) {
```

becomes

```ts
function start(code: number) {
```

that's it, little annotations on parameter types.
the rest of the changes are ESM related (not typescript specific)

> Any reason to not do the jsdoc approach?

https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html#null-undefined-and-empty-array-initializers-are-of-type-any-or-any
it problematic for micromark, the code style makes heavy use of uninitialized variables, nullish values, and empty untyped arrays.
Even when using JSDoc annotations.
An example

```js
// in JavaScript mode TS can't tell that this is meant to be a constant, it thinks it is meant to be an initialized export representing another currently unknown type.
exports.eof = null
```

wooorm:
> The difference in size doesn't necessarily have to do with TS.
> You re-architected the parser and compiler.

The difference in size may not necessarily be due to TS, but I’m assuming that lack of access to the output format will result in more bytes (this is also what I remember from coffeescript).
The re-architecting had to happen because the previous approach didn’t work.

> It's true that adding annotations technically makes it a superset/new language, it's not nearly as much of a learning curve as you make it out to be.

I’m afraid of it getting way more complex than a couple of types sprinkled on top (effects, okay, notokay). I foresee that once TS is in, the more powerful and more confusing things will also be used. Regarding the learning curve, I can only speak for myself and for what I’ve seen when others’ started using TS.

> It took hours to even partially reconstruct what a tokenizer can accept for this PR.
Most people probably wont be able to look at this code and understand what the undocumented, completed dynamic, inputs might be.

It is complex, there are no docs, and it needs to be better.

> the code style makes heavy use of uninitialized variables, nullish values, and empty untyped arrays

Null is only used for the EOF character. Much of the code style is written to work well with minifiers, manglers, and gzip.

ChristianMurphy:
> The difference in size may not necessarily be due to TS, but I’m assuming that lack of access to the output format will result in more bytes (this is also what I remember from coffeescript).

Not necessarily, again, it's JavaScript with annotations, remove the annotations and it's back to JavaScript.
Both the TypeScipt compiler and https://babeljs.io/docs/en/babel-preset-typescript can do this.

> foresee that once TS is in, the more powerful and more confusing things will also be used.

Using babel as the compiler gives pretty fine grained control over what features are/are not enabled.
https://babeljs.io/docs/en/babel-preset-typescript just removes annotations, any other language features need to be run through their respective babel plugins (and therefore can be disabled).
This also isn't limited to TypeScript, there are a lot of JavaScript features which could be used, but aren't, this seems like a slippery slope fallacy.

> > the code style makes heavy use of uninitialized variables, nullish values, and empty untyped arrays
>
> Null is only used for the EOF character. Much of the code style is written to work well with minifiers, manglers, and gzip.

Right, but that is just one example, the code is littered with `var something` and `var something = []`, uninitialized values, and untyped empty arrays.

lib/types.ts
```diff
@@ -0,0 +1,85 @@
+import type * as typesObject from './constant/types'
+import type {ValueOf} from 'type-fest'
+
+export interface Point {
+  line: number
+  column: number
+  offset: number
+  _index: number
+  _bufferIndex: number
+}
```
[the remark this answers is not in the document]
ChristianMurphy: are `_index` and `_buffer` optional or required?

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+export interface Token {
+  start: Point
+  end: Point
+  type: unknown
```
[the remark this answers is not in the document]
ChristianMurphy: What shape is the `type`?
Is it a `string`? If so are the possible values of the string known/enumerable (E.G. `list`, `link`, etc) or open ended?

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  contentType: string
```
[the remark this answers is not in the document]
ChristianMurphy: Does that mean this will be either `flow` or `content` as literal strings? Or is this more open ended?

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  previous: Token
+  next: Token
+  _size: number
```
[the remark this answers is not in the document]
ChristianMurphy: Is it always required?
Or will it only appear on certain token types?

lib/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  _break?: boolean
+  _tokenizer?: unknown
+  _events?: Event[]
+  _subevents?: Event[]
+  _contentTokenized?: unknown
```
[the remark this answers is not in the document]
ChristianMurphy:
```suggestion
  _contentTokenized?: boolean
```

lib/types.ts
ChristianMurphy: would this be `number | 'tab'` similar to remark?

lib/types.ts
ChristianMurphy: does consume not return a `Token`?

lib/types.ts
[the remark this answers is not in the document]
wooorm: Required, always there, set internally by the tokenizers (create-tokenizer).

lib/types.ts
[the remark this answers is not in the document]
wooorm: non-empty string. See https://github.com/micromark/micromark/blob/main/lib/constant/types.js. Although extensions will add other types.

lib/types.ts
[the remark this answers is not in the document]
wooorm: Correct, one of the names in the parser. As constants here: https://github.com/micromark/micromark/blob/8c8e92f1d05c480cabe91730136aada82cac8cdb/lib/constant/constants.js#L15-L18.

Extensions could potentially define more parsers, but I don’t really see that happening soon

lib/types.ts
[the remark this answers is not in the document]
wooorm: For gzip reasons, I use a lot of `true` values, where undefined/uninitialized is the other state.

This looks good.

lib/types.ts
wooorm: I think it’s an optional boolean

wooorm: (describes whether a token includes tabs, only used for trailing whitespace: https://github.com/micromark/micromark/blob/66f33dd1d3c22f7ed8ff0d12d38dd8cc5cba4165/lib/tokenize/whitespace.js)

lib/types.ts
wooorm: nope, returns void.

Enter and exit do return token.

lib/util/flat-map.ts
```diff
@@ -1,11 +1,9 @@
-module.exports = flatMap
-
 // Note: `a` could be given here, which is then passed to the map function.
 // It functions as a rest/spread, but smaller.
-function flatMap(array, map, a) {
+export default function flatMap<T, A>(array: T[][], map: (items: T[], a: A) => T[], a?: A): T[] {
```
ChristianMurphy, line 3: what is `a`? and how is it used?

wooorm: It was needed for (actual) buffers, as an encoding can be passed. In theory, it is anything that is given to `map` as the second param, after the value. In practise, it’s a Node buffer character encoding (such as `'utf8'`)

lib/types.ts
ChristianMurphy: Added some follow up questions in https://github.com/micromark/micromark/pull/17#discussion_r486744591  and https://github.com/micromark/micromark/pull/17#discussion_r486744731

lib/types.ts
ChristianMurphy: https://github.com/micromark/micromark/pull/17/files#r486745184

PR #17, opened 2020-09-10
types: add type definitions for micromark

ChristianMurphy:
alternative to #16

### DIFF
--- a/lib/compile/html.d.ts
+++ b/lib/compile/html.d.ts
@@ -1,0 +1,3 @@
+declare function compileHTML(): void
+
+export = compileHTML
--- a/shared-types.d.ts
+++ b/shared-types.d.ts
@@ -1,0 +1,220 @@
+// TypeScript Version: 3.0
+
+/**
+ * A location in a string or buffer
+ */
+export interface Point {
+  line: number
+  column: number
+  offset: number
+  _index?: number
+  _bufferIndex?: number
+}
+
+/**
+ * A token type
+ *
+ * TODO: enumerate token types
+ */
+export type Type = string
+
+/**
+ *
+ */
+export interface Token {
+  type: Type
+  start: Point
+  end: Point
+
+  previous: Token
+  next: Token
+
+  /**
+   * Declares a token as having content of a certain type.
+   * Because markdown requires to first parse containers, flow, content completely,
+   * and then later go on to phrasing and such, it needs to be declared somewhere on the tokens.
+   */
+  contentType: 'flow' | 'content' | 'string' | 'text'
+
+  // TODO move these to interfaces extending Token, or move these to a `data` property similar to unist (would require code refactor)
+  /**
+   * Used for whitespace in several places which needs to account for tab stops
+   */
+  _size?: number
+
+  /**
+   * ends with a CR, LF, or CRLF.
+   */
+  _break?: boolean
+
+  /**
+   * Used when dealing with linked tokens. A child tokenizer is needed to tokenize them, which is stored on those tokens
+   */
+  _tokenizer?: Tokenizer
+
+  /**
+   * Used for attention (emphasis, strong).
+   *
+   * could be (enter emphasis, enter emphasisMarker, exit emphasisMarker, enter strong, enter strongMarker, exit strongMarker, enter data, exit data)
+   */
+  _events?: Event[]
+
+  /**
+   * This is used for tokens that are already “subtokenized”.
+   *
+   * E.g., when parsing flow, there are content tokens, but those are directly tokenized into definitions/setext/paragraphs
+   */
+  _subevents?: Event[]
+
+  /**
+   * Set to true to mark that a token (e.g., with subevents) is already handled
+   */
+  _contentTokenized?: boolean
+
+  /**
+   * close and open are also used in attention:
+   * depending on the characters before and after sequences (**),
+   * the sequence can open, close, both, or none
+   */
+  _open?: boolean
+
+  /**
+   * close and open are also used in attention:
+   * depending on the characters before and after sequences (**),
+   * the sequence can open, close, both, or none
+   */
+  _close?: boolean
+  _marker?: number
+  _side?: number
+
+  /**
+   * Generally, tabs and spaces behave the same, but in the case of a hard break through trailing spaces ( \n), tabs
+   */
+  _tabs?: boolean
+}
+
+/**
+ *
+ */
+export type Event = [string, Token, unknown]
+
+/**
+ * These these are transitions to update the CommonMark State Machine (CSMS)
+ */
+export interface Effects {
+  /**
+   * Enter and exit define where tokens start and end
+   */
+  enter: (type: Type) => void
+
+  /**
+   * Enter and exit define where tokens start and end
+   */
+  exit: (type: Type) => void
+
+  /**
+   * Consume deals with a character, and moves to the next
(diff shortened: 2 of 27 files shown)


shared-types.d.ts
```diff
@@ -0,0 +1,220 @@
...
+   * If a value resulted in `ok`, it worked, the tokens that were made are used,
+   * and `returnState` is switched to.
+   * If the result is `nok`, the attempt failed,
+   * so we revert to the original state, and `bogusState` is used.
+   */
+  attempt: (
+    notSureWhatThisIs:
+      | Construct
+      | Construct[]
+      | {[code: number]: Construct | Construct[]},
```
ChristianMurphy, line 131: What is this construct like parameter called?

shared-types.d.ts
```diff
@@ -0,0 +1,220 @@
...
+  events: Event[]
+  parser: Parser
+  sliceStream: (token: Token) => string[]
+  sliceSerialize: (token: Token) => string
+  now: () => Point
+  defineSkip: (value: Point) => void
+  write: (value: number) => Event[]
+}
+
+export type Construct = unknown
```
ChristianMurphy, line 192: What is a construct?
Is it a function?

shared-types.d.ts
```diff
@@ -0,0 +1,220 @@
...
+    | Construct[]
+    | {[code: number]: Construct | Construct[]},
+  ok: Okay,
+  nok?: NotOkay) => void
+}
+
+/**
+ *
+ */
+export type Okay = (code: number) => () => void
```
ChristianMurphy, line 171: Does the inner function have any parameters?

meritozh:
I don't think syncing type definitions manually is a better way. `*.d.ts` should be generated by tsc.

ChristianMurphy:
@meritozh originally micromark was going to be in TypeScript (see https://github.com/remarkjs/remark/issues/439 and https://github.com/micromark/micromark/pull/4) however the direction changed. Bringing back native typescript was proposed again and declined https://github.com/micromark/micromark/pull/16#issuecomment-690326627.

This has also been discussed before in https://github.com/unifiedjs/unified/pull/76, most of the rest of unified also has typings files rather than being native TypeScript (see: https://github.com/search?p=2&q=org%3Asyntax-tree+typescript&type=Issues and https://github.com/search?q=org%3Aunifiedjs+typescript&type=Issues)
There is also a general propals to use JSDoc to run TypeScript https://github.com/unifiedjs/rfcs/pull/5, which is stalled at the moment.

If you want to move native TypeScript or jsdoc based TypeScript forward, starting a thread at https://github.com/micromark/micromark/discussions would be a good place to have the discussion.
Tacking the discussion on to this PR, is not a good place to restart that broad discussion.

shared-types.d.ts
wooorm: I dunno!

I’ve been calling the third form “hooks”, as it defines which character codes are “hooked”. But that doesn’t say too much.

Maybe something like “usable” in unified? Attempt?

shared-types.d.ts
wooorm: Should the params in check/lazy/interrupt also be called “returnState” and “bogusState”.

The tokenizer function gets `effects, ok, nok`, which is kind of like the executor function given to Promise: `resolve, reject`.
On the outside, you can pass `returnState, bogusState`, which is kind of like the functions given to `then`: `onfulfill, onreject`

They all get character codes

wooorm: All these functions also should return another function: the next state-as-a-function to go to.
But there is one case where they return void: for the eof character code (at the end of a value)
The reason being: well, there isn’t any state that makes sense, so void works well. Practically that has also helped: if for some reason it was a mistake, then an exception is throw because there is no next function, meaning it surfaces early.

shared-types.d.ts
wooorm: A construct is an object with:
- a tokenize function
- [resolve, resolveTo](https://github.com/micromark/micromark/blob/005ad17216ba591560a1a73026c35d2a669fc2a7/lib/util/create-tokenizer.js#L332-L340), and [resolveAll](https://github.com/micromark/micromark/blob/005ad17216ba591560a1a73026c35d2a669fc2a7/lib/util/create-tokenizer.js#L77) functions (which take a certain slice of events and can mutate them)
- [`partial`](https://github.com/micromark/micromark/blob/005ad17216ba591560a1a73026c35d2a669fc2a7/lib/util/create-tokenizer.js#L289) (bit ugly, used for things that aren’t really constructs, like whitespace)
- [`concrete`](https://github.com/micromark/micromark/blob/005ad17216ba591560a1a73026c35d2a669fc2a7/lib/tokenize/html-flow.js#L3), used for multiline constructs such as HTML and fenced code that and how they mix with containers (e.g., `> <div>\n> > this is html, not a second block quote`)
- [Content has a name](https://github.com/micromark/micromark/blob/005ad17216ba591560a1a73026c35d2a669fc2a7/lib/initialize/flow.js#L16)

shared-types.d.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Enter and exit define where tokens start and end
+   */
+  exit: (type: Type) => void
+
+  /**
+   * Consume deals with a character, and moves to the next
+   */
+  consume: (code: number) => void
```
wooorm, line 118: Would it be a good idea to make a specific type for the character codes?

It can be any result (positive int) from String.charCodeAt (except NaN). And the following special codes are used:
https://github.com/micromark/micromark/blob/005ad17216ba591560a1a73026c35d2a669fc2a7/lib/character/codes.js#L1-L7

shared-types.d.ts
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * A location in a string or buffer
+ */
+export interface Point {
+  line: number
+  column: number
+  offset: number
+  _index?: number
+  _bufferIndex?: number
```
wooorm, line 11: For the underscore props, as those are kinda for internal use in this project, should they be left undocumented?

Is there some comment to add here to say to TS that it shouldn’t be exported to library users?

lib/compile/html.d.ts
```diff
@@ -0,0 +1,3 @@
+declare function compileHTML(): void
```
wooorm, line 1: `compileHtml`?

shared-types.d.ts
ChristianMurphy: We could tag these with tsdoc
```
/**
 * @internal
 */
```
I'm cautious about removing them entirely, since extensions/plugins may leverage some of the internal attributes.

shared-types.d.ts
ChristianMurphy: It could be a good idea.
Both the known types and character codes could be turned in a union of known types or turned into an [opaque type](https://codemix.com/opaque-types-in-javascript), currently both are transparent primitives.

ocavue: If micromark is written in native TypeScript, we can use TypeScript's `const enum` feature for these constant values. 

By using `const enum`, those import statements while compiling and we don't need to use `script/babel-transform-constants.js` to do that. And we have an easy way to define types for character codes.

However, TypeScript doesn't allow `null` as a value of `const enum`, so we have to do something special for `eof`.

ChristianMurphy:
closing in favor of #25 which builds on this

PR #18, opened 2020-09-17
test: automatic fuzz and corpus generating test

ChristianMurphy:
This uses https://gitlab.com/gitlab-org/security-products/analyzers/fuzzers/jsfuzz to automatically generate test cases checking for edge cases where micromark would throw an exception instead of returning text.


fuzzer.js
```diff
@@ -0,0 +1,16 @@
+const micromark = require('./index')
+
+function fuzz(buf) {
+  try {
+    micromark(buf)
+  } catch (e) {
+    if (e.message.indexOf('URI malformed') !== -1) {
```
ChristianMurphy, line 7: Ignore `URI malformed` errors that can bubble up from the `URI` module.

package.json
```diff
@@ -85,6 +86,7 @@
     "format": "remark . -qfo && prettier . --write && xo --fix",
     "test-api": "node test",
     "test-coverage": "nyc --reporter lcov tape test/index.js",
+    "test-fuzz": "jsfuzz fuzzer.js corpus",
```
ChristianMurphy, line 89: :notebook: the fuzzer can run forever, it's probably best not to run the fuzzer itself on CI.

However, the corpus it generates could potentially be used for testing.
Including the corpus for testing is optional, in this PR I did not include any of the generated test cases (yet).

ChristianMurphy: It seems that the fuzzer can explore up to around 90% coverage in an hour or two.
It may be worth exploring the possibility of occasionally running the fuzzer for a couple hours the shutting it down based off a timer, to explore edge cases potentially not covered in the unit tests.

wooorm: Travis seems to have a run time of 50 minutes tops. Could we perhaps have this in a separate repo (micromark/fuzzcheck or so), with a weekly/monthly travis cronjob? 🤔

wooorm: Are there other things like this we could (automatically) test for?

ChristianMurphy: :thinking: a few other things:
* [Symbolic Execution of the code](https://en.wikipedia.org/wiki/Symbolic_execution) (example [1](https://github.com/Samsung/jalangi2), [2](https://github.com/ExpoSEJS/ExpoSE)), like fuzzing, SymEx can be very time consuming.
* [More ESLint rules](https://github.com/dustinspecker/awesome-eslint#code-quality), like [sonar](https://github.com/SonarSource/eslint-plugin-sonarjs), [unicorn](https://github.com/sindresorhus/eslint-plugin-unicorn), [node](https://github.com/mysticatea/eslint-plugin-node), and [regex](https://github.com/BrainMaestro/eslint-plugin-optimize-regex)

ChristianMurphy: > Travis seems to have a run time of 50 minutes tops. Could we perhaps have this in a separate repo (micromark/fuzzcheck or so), with a weekly/monthly travis cronjob?

That's one option, another would be to leverage GithHub actions, which have a limit of 6 hours.
https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits

fuzzer.js
```diff
@@ -0,0 +1,52 @@
...
+      htmlExtensions: [gfmHtml]
+    })
+
+    // gfm buffer with html
+    micromark(buf, 'utf-8', {
+      allowDangerousHtml: true,
+      allowDangerousProtocol: true,
+      extensions: [gfmSyntax()],
+      htmlExtensions: [gfmHtml]
+    })
```
ChristianMurphy, line 40: these could potentially be moved the extension repo rather than on micromark itself. :man_shrugging:

package.json
wooorm: damn that’s long 😅 well. All for doing that. I think that having it in some cron job will help because otherwise we’ll forget. But on the other hand: someone has to check when the cron job fails. Does GH CI have cron jobs, and some notification if it fails?

lint: I’m all for more rules, although a) xo is very stylistic (to the point of being wrong when doing such lowlevel things), b) I think coverage will have checked sonar and regex. But I’m all for trying ’em out!

CST + fuzz: ah, yeah, that would be good! I’m now working on the AST, and there’s just like: so many ways for markdown to accidentally parse differently. Same with HTML: it also can’t round trip 100%. I think that’s fine for the AST. Although it makes sense to me that the CST should roundtrip 100%.
To illustrate the AST: GFM parses `a.b-c_d@a.b` as an email link. mdast-util-to-markdown serializes that as `<a.b-c_d@a.b>`, so that it is also a link to non-GFM. However, CM has slightly different email parsing because `a` is too short for a domain name so it bails. The only way to properly address these edge cases is to serialize, use a markdown parser to parse the result, and see if it’s indeed the same as the origin, and that’s just too much work.

ChristianMurphy: > But on the other hand: someone has to check when the cron job fails. Does GH CI have cron jobs, and some notification if it fails?

it does support scheduled/cron jobs https://docs.github.com/en/actions/reference/events-that-trigger-workflows#schedule
for notifications, I'm not sure what the default behavior is, but there are actions to sent emails, slack messages, etc.

wooorm: ok, let’s first focus on getting this in, then figure out where we’re running cron jobs!

lib/constant/unicode-punctuation-regex.js
```diff
@@ -4,4 +4,4 @@
...
...055F\u0589\u058A\u05BE\u05C0\u05C3\u05C6\u05F3\u05F4\u0609\u060A\u060C\u060D\u061B\u061E\u061F\u066A-\u066D\u06D4\u0700-\u070D\u07F7-\u07F9\u0830-\u083E\u085E\u0964\u0965\u0970\u09FD\u0A76\u0AF0\u0C77\u0C84\u0DF4\u0E4F\u0E5A\u0E5B\u0F04-\u0F12\u0F14\u0F3A-\u0F3D\u0F85\u0FD0-\u0FD4\u0FD9\u0FDA\u104A-\u104F\u10FB\u1360-\u1368\u1400\u166E\u169B\u169C\u16EB-\u16ED\u1735\u1736\u17D4-\u17D6\u17D8-\u17DA\u1800-\u180A\u1944\u1945\u1A1E\u1A1F\u1AA0-\u1AA6\u1AA8-\u1AAD\u1B5A-\u1B60\u1BFC-\u1BFF\u1C3B-\u1C3F\u1C7E\u1C7F\u1CC0-\u1CC7\u1CD3\u2010-\u2027\u2030-\u2043\u2045-\u2051\u2053-\u205E\u207D\u207E\u208D\u208E\u2308-\u230B\u2329\u232A\u2768-\u2775\u27C5\u27C6\u27E6-\u27EF\u2983-\u2998\u29D8-\u29DB\u29FC\u29FD\u2CF9-\u2CFC\u2CFE\u2CFF\u2D70\u2E00-\u2E2E\u2E30-\u2E4F\u2E52\u3001-\u3003\u3008-\u3011\u3014-\u301F\u3030\u303D\u30A0\u30FB\uA4FE\uA4FF\uA60D-\uA60F\uA673\uA67E\uA6F2-\uA6F7\uA874-\uA877\uA8CE\uA8CF\uA8F8-\uA8FA\uA8FC\uA92E\uA92F\uA95F\uA9C1-\uA9CD\uA9DE\uA9DF\uAA5C-\uAA5F\uAADE\uAADF\uAAF0\uAAF1\uABEB\uFD3E\uFD3F\uFE10-\uFE19\uFE30-\uFE52\uFE54-\uFE61\uFE63\uFE68\uFE6A\uFE6B\uFF01-\uFF03\uFF05-\uFF0A\uFF0C-\uFF0F\uFF1A\uFF1B\uFF1F\uFF20\uFF3B-\uFF3D\uFF3F\uFF5B\uFF5D\uFF5F-\uFF65]/
```
wooorm, line 7: npm test would reformat this back

package.json
```diff
@@ -68,8 +68,12 @@
     "eslint-plugin-security": "^1.0.0",
     "eslint-plugin-sonarjs": "^0.5.0",
     "gzip-size-cli": "^3.0.0",
+    "jsfuzz": "1.0.14",
+    "micromark-extension-frontmatter": "^0.2.0",
+    "micromark-extension-gfm": "^0.2.0",
```
wooorm, line 73:
```suggestion
    "micromark-extension-gfm": "^0.3.0",
```

wooorm (commented):
I’m not sure about pulling all the fixtures into this repo. Could we create a corpus `micromark/lots-of-markdown` or so? Does one big file work, or should it be a directory of lots of files?

wooorm:
I landed a version of this which credits you, @ChristianMurphy.

I checked this out but kept running into problems where jsfuzz made 90mb markdown files based on the remark fixtures (which are sometimes large). Which seems over a sensible limit to micromark. So, I instead opted to generate fixtures from the tests by capturing all the strings passed.

Interestingly, I’ve had more luck removing the fixtures directory, which jsfuzz uses to store intermediate files too, capturing the fixtures again, and then fuzzing again: clean runs of jsfuzz seem to find bugs fast, whereas reruns seem to not really produce anything. Hence, I haven’t checked the fixtures in.

Two future things to add later are a) cron job, b) making this somehow usable for extensions.

And, to add re https://github.com/micromark/micromark/pull/18#discussion_r491081473, I did try out more eslint rules, which [caught one thing](https://github.com/micromark/micromark/commit/f0d8bb35529de46cba0ec668cdd0b3ce19e8c280#diff-c5cade5961d1e67899cca502ac3dae1f).

Issue #20, opened 2020-09-19
unravelLinkedTokens RangeError: Maximum call stack size exceeded


ChristianMurphy:
### Subject of the issue

#18 fed with https://github.com/remarkjs/remark/blob/8108fe54e04640dda119aad366d70e6edf2602f1/test/fixtures/input/title-attributes.text can trigger a call stack exceeded issue in `unravelLinkedTokens`.
These files are pretty large 1mb and around 30k lines a piece, a more minimal example, at 105kb is also included.

It seems the be related to unterminated links, but more research is needed.

### Your environment

*   **OS**: Ubuntu
*   **Packages**: micromark 2.6.1
*   **Env**: node 14

### Steps to reproduce

```js
var fs = require('fs')
var micromark = require('./index')

// var doc = fs.readFileSync('crash-395a731d55c510f1338b8c9911c159ab56329d18bc3a12a26b826b750d0b1253.txt')
// var doc = fs.readFileSync('crash-4bf6a4882505b11dea88b5e16e6f0d3766252601ae704e42ebe606d270f9f26f.txt')
var doc = fs.readFileSync('crash-7182fa3e89e1b8fb28bda27b6da6b3769f05b1ce68551d96c46acd0931d95004.txt')

var result = micromark(doc)

console.log(result)
```

[crash-7182fa3e89e1b8fb28bda27b6da6b3769f05b1ce68551d96c46acd0931d95004.txt](https://github.com/micromark/micromark/files/5248998/crash-7182fa3e89e1b8fb28bda27b6da6b3769f05b1ce68551d96c46acd0931d95004.txt)
[crash-4bf6a4882505b11dea88b5e16e6f0d3766252601ae704e42ebe606d270f9f26f.txt](https://github.com/micromark/micromark/files/5248999/crash-4bf6a4882505b11dea88b5e16e6f0d3766252601ae704e42ebe606d270f9f26f.txt)
[crash-395a731d55c510f1338b8c9911c159ab56329d18bc3a12a26b826b750d0b1253.txt](https://github.com/micromark/micromark/files/5249009/crash-395a731d55c510f1338b8c9911c159ab56329d18bc3a12a26b826b750d0b1253.txt)

a more minimal example of what may be the same issue (`[](` repeated 35k times in a 105kb file)

[repeated-unterminated-links.txt](https://github.com/micromark/micromark/files/5249068/nasty-links.txt)

### Expected behavior

If possible no error, alternatively a better error message could help.

### Actual behavior

```
[...]

PR #21, opened 2020-09-22
Fix stackoverflow caused by unravelLinkedTokens and Array#splice

fazouane-marouane:
1. This fixes the three crash reports from https://github.com/micromark/micromark/issues/20 which is related to unravelLinkedTokens
2. Array#splice takes all items to be inserted as individual argument which also causes a stackoverflow


lib/constant/splice.js
```diff
@@ -1 +0,0 @@
-module.exports = [].splice
```
wooorm, line 1: can we keep this one and… (1)

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Plain old regular Array#splice may cause stackoverflow if we have too many items
+ */
+function chunkedSplice(array, start, deleteCount, items) {
+  var chunkSize = 10000
+  var splice = [].splice
```
wooorm, line 6: (1) …use it for this one?

Some of the code is written in a funky way as it improves gzip a bunch, having these constant methods has helped shave of bytes

lib/util/subtokenize.js
```diff
@@ -122,7 +122,7 @@ function subtokenize(events) {
   return {done: !more, events: events}
 }
 
-function unravelLinkedTokens(token, context) {
+function unravelLinkedTokensImpl(token, context) {
```
ChristianMurphy, line 125: This handles a single token, maybe

```suggestion
function unravelLinkedToken(token, context) {
```

or

```suggestion
function unravelLinkedTokenImplementation(token, context) {
```

would make sense?

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Plain old regular Array#splice may cause stackoverflow if we have too many items
+ */
+function chunkedSplice(array, start, deleteCount, items) {
+  var chunkSize = 10000
```
wooorm, line 5: Move this one a scope up

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * Plain old regular Array#splice may cause stackoverflow if we have too many items
+ */
```
wooorm, line 3: Can you add some more comments here about your findings? (also: maybe call this module `chunked-splice`; and could you use `//` comments?)

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
...
+function chunkedSplice(array, start, deleteCount, items) {
+  var chunkSize = 10000
+  var splice = [].splice
+
+  // Make deleteCount >= 0
+  deleteCount = Math.max(deleteCount, 0)
```
wooorm, line 9:
```suggestion
  deleteCount = deleteCount > 0 ? deleteCount : 0
```

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
...
+  // Make start between 0 and array.length (included)
+  start = Math.min(start, array.length)
+  if (start < 0) {
+    start = Math.max(array.length + start, 0)
+  }
```
wooorm, line 14: (I believe the first case works)

```suggestion
  if (start < 0) {
    start = -start > array.length ? 0 : array.length + start
  } else {
      start = start > array.length ? array.length : start
  }
```

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (items.length < chunkSize) {
+    // No need to chunk the items in this case
+    return splice.apply(array, [start, deleteCount].concat(items))
```
wooorm, line 18: Will mutating help perf here?

```suggestion
    items.unshift(start, deleteCount)
    return splice.apply(array, items)
```

wooorm: Might be bad to mutate, but as this is such an internal function which is called a bazillion times, it might make sense to do funky things like that if it helps

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  // Delete deleteCount items starting from `start`
+  var result = splice.apply(array, [start, deleteCount])
+  // Insert the items by chunks so we won't cause a stackoverflow
+  var chunkStart = 0
+  while (chunkStart < items.length) {
+    var chunk = items.slice(chunkStart, chunkStart + chunkSize)
```
wooorm, line 26:
```suggestion
    var chunk = items.slice(chunkStart, chunkStart + chunkSize)
    chunk.unshift(start, 0)
```

(same as above)

lib/util/subtokenize.js
```diff
@@ -155,13 +155,21 @@ function unravelLinkedTokens(token, context) {
...
-    divideTokens(token)
+    if (token.next) {
+      // Unravel the next token.
+      token = token.next
+    } else {
+      // Done!
+      divideTokens(token)
+      return
+    }
   }
```
wooorm, line 173:
```suggestion
    // Unravel the next token.
    token = token.next
  }

  // Done!
  divideTokens(token)
}
```

This should work I believe?

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
...
+    start = Math.max(array.length + start, 0)
+  }
+
+  if (items.length < chunkSize) {
+    // No need to chunk the items in this case
+    return splice.apply(array, [start, deleteCount].concat(items))
+  }
+
+  // Delete deleteCount items starting from `start`
+  var result = splice.apply(array, [start, deleteCount])
```
wooorm, line 22: Can you move the variable declarations to the top of the function?

wooorm (commented):
Couple style nits, notes on size, and maybe some further performance enhancements?

lib/util/splice.js
wooorm: Also: would it perhaps be better for performance to directly copy `start..deleteCount` in from `items` (assuming `deleteCount < items.length`)?

lib/util/subtokenize.js
fazouane-marouane: agreed !

lib/util/splice.js
```diff
@@ -0,0 +1,36 @@
...
+  // Insert the items by chunks so we won't cause a stackoverflow
+  var chunkStart = 0
+  while (chunkStart < items.length) {
+    var chunk = items.slice(chunkStart, chunkStart + chunkSize)
+    splice.apply(array, [start, 0].concat(chunk))
```
fazouane-marouane, line 27:
```suggestion
    splice.apply(array, chunk)
```

lib/util/splice.js
fazouane-marouane: Looks good indeed !

lib/util/splice.js
fazouane-marouane, line 3:
```suggestion
var chunkSize = 10000

// Plain old regular Array#splice may cause stackoverflow if we have too many items
```

lib/util/splice.js
fazouane-marouane, line 5:
```suggestion
```

fazouane-marouane: fixed

lib/util/splice.js
fazouane-marouane: fixed

lib/constant/splice.js
fazouane-marouane: fixed

lib/util/subtokenize.js
fazouane-marouane: It seems to be broken since token would be undefined but we still need to pass it to `divideTokens`

ChristianMurphy (approved):

lib/util/splice.js
fazouane-marouane: mutation might be tricky since we may introduce bugs that are hard to debug. Example from the current commit:

```js
chunkedSplice(events, index, 2, subevents)
index += subevents.length - 1
```

wooorm (approved):
Let’s try this out!

PR #25, opened 2020-10-03
Add types

ocavue:
Add more type definitions based on the awesome work from @ChristianMurphy at #17


shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * Generally, tabs and spaces behave the same, but in the case of a hard break through trailing spaces ( \n), tabs
+   */
+  _tabs?: boolean
+}
+
+/**
+ *
+ */
+export type Event = [string, Token, unknown]
```
wooorm, line 99: unknown is a TokenizerContext

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+   * If the result is `nok`, the attempt failed,
+   * so we revert to the original state, and `bogusState` is used.
+   */
+  attempt: (
+    constructInfo:
+      | Construct
+      | Construct[]
+      | { [code: number]: Construct | Construct[] },
+    returnState: unknown,
+    bogusState?: unknown
```
wooorm, line 133: the two unknowns are `State`s

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+/**
+ *
+ */
+export type NotOkay = State
+
+/**
+ *
+ */
+export interface Tokenizer {
+  previous: Token
```
wooorm, line 194: previous is a `Code`

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+  events: Event[]
+  parser: Parser
+  sliceStream: (token: Token) => string[]
+  sliceSerialize: (token: Token) => string
+  now: () => Point
+  defineSkip: (value: Point) => void
+  write: (value: number) => Event[]
```
wooorm, line 201: write receives Chunks

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+export type NotOkay = State
+
+/**
+ *
+ */
+export interface Tokenizer {
+  previous: Token
+  events: Event[]
+  parser: Parser
+  sliceStream: (token: Token) => string[]
```
wooorm, line 197: sliceStream results in Chunks

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+export type Tokenize = (context: unknown, effects: Effects) => State
+
+export interface Construct {
+  tokenize: Tokenize
+  partial?: boolean
+  resolve?: Resolve
+  resolveTo?: Resolve
+  resolveAll?: Resolve
+  concrete?: boolean
+  name: string
```
wooorm, line 215: I believe `name` is removed, and these two added:

https://github.com/micromark/micromark/blob/027d599907f0f5b12b603158a01215594b87ad4c/lib/tokenize/content.js#L3-L4

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ *
+ */
+export interface Parser {
+  hooks: {
+    [key: string]: Construct | Construct[]
+  }
+  flow: (something: unknown) => unknown
+  defined: unknown[]
```
wooorm, line 226: Array of strings

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+  concrete?: boolean
+  name: string
+}
+
+/**
+ *
+ */
+export interface Parser {
+  hooks: {
+    [key: string]: Construct | Construct[]
```
wooorm, line 223:
```suggestion
    [key: Code]: Construct[]
```

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+  resolveAll?: Resolve
+  concrete?: boolean
+  name: string
+}
+
+/**
+ *
+ */
+export interface Parser {
+  hooks: {
```
wooorm, line 222:
```suggestion
  constructs: {
```

shared-types.d.ts
```diff
@@ -0,0 +1,269 @@
...
+    [key: string]: Construct | Construct[]
+  }
+  flow: (something: unknown) => unknown
```
wooorm, line 225: maybe add the others too: https://github.com/micromark/micromark/blob/027d599907f0f5b12b603158a01215594b87ad4c/lib/parse.js#L19-L23

They’re calls to createTokenizer, so they return a tokenizer

wooorm (commented):
Awesome, thanks @ocavue!

lib/util/shallow.d.ts
```diff
@@ -0,0 +1,3 @@
+declare function shallow<T>(object: T): T
```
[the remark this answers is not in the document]
ocavue: A plain-vanilla object can still have typing information, which could be helpful.

For example:

```typescript
interface Point {
  line: number
  column: number
  offset: number
}

function createTokenizer(from?: Point) {
  var point = from ? shallow(from) : {line: 1, column: 1, offset: 0}
  // ... 
}
```

If we keep the current definition, TS can infer that the data type of `point` is `Point`.

lib/util/serialize-chunks.d.ts
```diff
@@ -0,0 +1,3 @@
+declare function serializeChunks(chunks: Array<string | number>): string
```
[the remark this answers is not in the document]
ocavue: Does `Chunk` has a specific definition like `Code` or it can be any `string | number`? 

According to [this function](https://github.com/micromark/micromark/blob/2.9.2/lib/preprocess.js#L52), it seems that `Chunk` is a subset of `Code` but I'm not sure.

shared-types.d.ts
ocavue: TypeScript will complain that `null` can't be a key of an object. Is `null` (or `"null"`) a possible key of `hooks`?

lib/util/classify-character.d.ts
```diff
@@ -0,0 +1,3 @@
+declare function classifyCharacter(code: number): number
```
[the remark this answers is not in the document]
ocavue: I added `Codes` in `lib/character/codes.d.ts` and `Code` in `shared-types.d.ts`

lib/index.d.ts
```diff
@@ -0,0 +1,7 @@
+import {ParserOptions} from './parse'
+
+declare function buffer(value: string | Buffer, options?: ParserOptions): string
+
+declare function buffer(value: string | Buffer, encoding?: BufferEncoding, options?: ParserOptions): string
```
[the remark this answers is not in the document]
ocavue: > Is `BufferEncoding` injected by TS because this is defined as a Node module?

Yes. `BufferEncoding` is from `node_modules/@types/node/` 

> If so: in `vfile` we opted to copy the list instead:

It easy to copy `BufferEncoding`, but it's not easy to copy [`Buffer`](https://github.com/ocavue/micromark/blob/6a53e9486c0ba1e9faed3d70ebc03f162d410f87/lib/index.d.ts#L6), which is also from `@types/node` but more complex. Maybe we should add `"@types/node": ">=10.0.0"` as one of the dependencies?

micromark.test.ts
```diff
@@ -0,0 +1,5 @@
+import micromarkBuffer = require('micromark')
+import micromarkStream = require('micromark/stream')
+
+micromarkBuffer("")
+micromarkStream()
```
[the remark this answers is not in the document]
ocavue: Maybe we can write test files into native TypeScript. By doing that, we can test the logic and typing together. And we can also avoid some shortcomings about TypeScript like "lack of access to the output format". Do you think it's a good idea?

buffer.d.ts
```diff
@@ -0,0 +1,5 @@
+// TypeScript Version: 3.0
+
+import buffer = require('./lib')
```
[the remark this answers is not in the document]
ocavue: If we point this to `dist/`, we have to copy our `.d.ts` files into `dist/` too, which seems unnecessary. Thus, I think the current way is better.

package.json
```diff
@@ -60,11 +60,13 @@
   "devDependencies": {
```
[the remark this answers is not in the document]
ocavue: > package.json should also have a types field.

fixed.

> Most of our types are not in the main code, but in a types/ folder. Should we do that here too? Or does this make more sense?

I think `lib/` is a better choice for `micromark`. Because `micromark` has a relatively complex directory structure and it's easy to have problems if we maintain the same directory structure in both `lib/` and `types/`.

micromark.test.ts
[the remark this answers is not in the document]
wooorm: I’m fine with some tests in typescript, but I want the main code (including build scripts and tests) to be JavaScript. I want someone who doesn’t know TS to be able to write a PR that comes with a test.

package.json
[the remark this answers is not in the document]
wooorm: Hmm. I still feel weird about having them in `lib/` as that is published only for debugging. Everybody is using the `dist/` code.

* I see your point about having the definitions close to the source, so writing/maintaining types in `lib/` makes sense
* Publishing the types in both `lib/` and `dist/` is superfluous
* In the published package, e.g., when debugging your node_modules, I think it makes more sense for them to live in `dist/`.

From these three points, could we have the source in `lib/`, copy them into `dist/` when building, and then in `package.json` ignore the ones in `lib/` but publish the ones in `dist/`?

shared-types.d.ts
```diff
@@ -0,0 +1,274 @@
...
+  column: number
+  offset: number
+  _index?: number
+  _bufferIndex?: number
+}
+
+/**
+ * A token type
+ */
+export type Type = Types[keyof Types]
```
wooorm, line 20: The goal is also to have types extendable btw, e.g., `micromark-extension-gfm` adds other types.

Would that work with this?

lib/util/serialize-chunks.d.ts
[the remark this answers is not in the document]
wooorm: It is an array of Code or string. The EOF (null) code never occurs!

lib/constant/unicode-punctuation-regex.js
```diff
@@ -4,4 +4,4 @@
...
...055F\u0589\u058A\u05BE\u05C0\u05C3\u05C6\u05F3\u05F4\u0609\u060A\u060C\u060D\u061B\u061E\u061F\u066A-\u066D\u06D4\u0700-\u070D\u07F7-\u07F9\u0830-\u083E\u085E\u0964\u0965\u0970\u09FD\u0A76\u0AF0\u0C77\u0C84\u0DF4\u0E4F\u0E5A\u0E5B\u0F04-\u0F12\u0F14\u0F3A-\u0F3D\u0F85\u0FD0-\u0FD4\u0FD9\u0FDA\u104A-\u104F\u10FB\u1360-\u1368\u1400\u166E\u169B\u169C\u16EB-\u16ED\u1735\u1736\u17D4-\u17D6\u17D8-\u17DA\u1800-\u180A\u1944\u1945\u1A1E\u1A1F\u1AA0-\u1AA6\u1AA8-\u1AAD\u1B5A-\u1B60\u1BFC-\u1BFF\u1C3B-\u1C3F\u1C7E\u1C7F\u1CC0-\u1CC7\u1CD3\u2010-\u2027\u2030-\u2043\u2045-\u2051\u2053-\u205E\u207D\u207E\u208D\u208E\u2308-\u230B\u2329\u232A\u2768-\u2775\u27C5\u27C6\u27E6-\u27EF\u2983-\u2998\u29D8-\u29DB\u29FC\u29FD\u2CF9-\u2CFC\u2CFE\u2CFF\u2D70\u2E00-\u2E2E\u2E30-\u2E4F\u2E52\u3001-\u3003\u3008-\u3011\u3014-\u301F\u3030\u303D\u30A0\u30FB\uA4FE\uA4FF\uA60D-\uA60F\uA673\uA67E\uA6F2-\uA6F7\uA874-\uA877\uA8CE\uA8CF\uA8F8-\uA8FA\uA8FC\uA92E\uA92F\uA95F\uA9C1-\uA9CD\uA9DE\uA9DF\uAA5C-\uAA5F\uAADE\uAADF\uAAF0\uAAF1\uABEB\uFD3E\uFD3F\uFE10-\uFE19\uFE30-\uFE52\uFE54-\uFE61\uFE63\uFE68\uFE6A\uFE6B\uFF01-\uFF03\uFF05-\uFF0A\uFF0C-\uFF0F\uFF1A\uFF1B\uFF1F\uFF20\uFF3B-\uFF3D\uFF3F\uFF5B\uFF5D\uFF5F-\uFF65]/
```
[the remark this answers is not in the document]
ocavue: I run `npm test` and has `Error: Failed to load plugin 'prettier' declared in 'CLIOptions': Cannot find module 'eslint-plugin-prettier'` error. Is it expected?

shared-types.d.ts
```diff
@@ -0,0 +1,274 @@
...
+  concrete?: boolean
+  interruptible?: boolean
+  lazy?: boolean
+}
+
+/**
+ *
+ */
+export interface Parser {
+  constructs: Record<NonNullable<Code>, Construct | Construct[]>
```
wooorm, line 187: To continue the conversation: https://github.com/micromark/micromark/pull/25#discussion_r499209678

It is indeed a `String(code)`, which is done implicitly when setting `null` or `number` on an object by JavaScript.
The key EOF code `null` is thus `"null"`

lib/index.d.ts
```diff
@@ -0,0 +1,11 @@
+import {ParseOptions} from '../shared-types'
+
+declare function buffer(value: string | Buffer, options?: ParseOptions): string
+
+declare function buffer(
+  value: string | Buffer,
+  encoding?: BufferEncoding,
```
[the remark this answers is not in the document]
ocavue: That makes sense.

shared-types.d.ts
ocavue: I will do some research on extendable types.

lib/constant/unicode-punctuation-regex.js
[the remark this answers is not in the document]
wooorm: I think that’s yarn having problems, I’ve heard about it before. Everything’s fine on npm.

shared-types.d.ts
ocavue: changed the type here to `NonNullable<Code> | 'null'`

shared-types.d.ts
ocavue: > From these three points, could we have the source in lib/, copy them into dist/ when building, and then in package.json ignore the ones in lib/ but publish the ones in dist/?

This sounds ok to me. I will do this later.

lib/character/codes.d.ts
```diff
@@ -0,0 +1,140 @@
+// This module is generated by `script/`.
```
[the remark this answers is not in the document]
ocavue: since we currently don't have a "compile" step for `.d.ts` (we could have "copy" step for them), I'm not sure what do you mean by "compiled away".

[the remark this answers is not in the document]
wooorm: These constants are inlined: https://github.com/micromark/micromark/blob/main/script/babel-transform-constants.js. They’re not used in the code in `dist/`

shared-types.d.ts
wooorm: Here are the ones for mdast and hast, which might serve as inspiration: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mdast/index.d.ts
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/hast/index.d.ts

It doesn’t look like they do anything for extending tho

ocavue: Maybe we can make `buffer` a function with two generic types: `SE extends SyntaxExtension`, `EE extends HtmlExtension` so that we can pass the extra `Type` into it.

something like this:

```
declare function buffer<
  SE extends SyntaxExtension,
  EE extends HtmlExtension,
  >(
    value: string | Buffer,
    encoding?: BufferEncoding,
    options?: ParseOptions<SE, EE>
  ): string
```
 

But this is too much modification: every function that accepts `Token` will need to add generic. So for now, let's just set  `Type` as `string` to avoid errors.

wooorm: Sounds good to me! @ChristianMurphy might have ideas though

lib/character/codes.d.ts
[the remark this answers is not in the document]
ocavue: I get your idea. Since the `.d.ts` files for all compiled away files are generated (see `script/generate-constant-typings.js`), I can generate something like this 

```ts
// lib/character/codes.d.ts
// This module is generated by `script/`.

export type Code = -5 | -4 | -3

// @script: REMOVE_ALL_THING_BELOW

export interface Codes {
  "carriageReturn": -5,
  "lineFeed": -4,
  "carriageReturnLineFeed": -3,
}

declare const value: Codes
export default value

// @script: REMOVE_ALL_THING_ABOVE
```

and remove all content between two comments when coping `*.d.ts` from `lib/` to `dist/`. By doing that, `dist/character/codes.d.ts` will only export a *type* without any *value*. 

There is several ways to do that, but this way is the fastest to implement.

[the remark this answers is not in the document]
ocavue: update: implemented.

lib/shared-types.d.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+  | {
+      exit: unknown
+    }
+
+export type ParseOptions = {
+  // Array of syntax extensions
+  //
+  extensions?: SyntaxExtension[]
+} & CompileOptions
```
wooorm, line 239: Should there be a `ParseAndCompileOptions` that includes both? Or just one called `Options`? So `Options` are given by users to buffer and stream, but `lib/parse` only takes the parse options, and `lib/compiler/html` takes the compile options?

script/copy-dict.js
```diff
@@ -0,0 +1,48 @@
...
+    else if (stat.isFile()) yield entry
+  }
+}
+
+function copy(libFilePath) {
+  var distFilePath = path.join(distDir, path.relative(libDir, libFilePath))
+  var dir = path.dirname(distFilePath)
+  if (!fs.existsSync(dir)) {
+    fs.mkdirSync(dir, {recursive: true}) // Need Node.js 10.12.0+
+  }
```
wooorm, line 24:
```suggestion
  fs.mkdirSync(dir, {recursive: true}) // Need Node.js 10.12.0+
```

I don’t think it’s needed to check?

script/copy-dict.js
```diff
@@ -0,0 +1,48 @@
...
+    outputLines.push(line)
+  }
+
+  fs.writeFileSync(distFilePath, outputLines.join('\n'))
+}
+
+function main() {
+  for (var filePath of walk(libDir)) {
+    if (filePath.endsWith('.d.ts')) {
+      console.log('copying', filePath)
```
wooorm, line 42: that’s probably a bit verbose?

```suggestion
```

script/generate-constant-typings.js
```diff
@@ -0,0 +1,41 @@
+// This script will read constants from `.js` files and generate `.d.ts` files.
+
+var fs = require('fs')
+var path = require('path')
+
+function generateTyping(name, paths) {
+  var constants = JSON.parse(JSON.stringify(require('../' + paths.join('/'))))
```
wooorm, line 7:
```suggestion
  var constants = require('../' + paths.join('/'))
```

Any reason for it to be deep cloned?

shared-types.d.ts
ChristianMurphy: :thinking: I don't think we'll ever remove types? Only add them?
If so, something like an interestion `type Type<Extras> = Types[keyof Types] & Extras`.
Then as @ocavue suggests those types would likely bubble up to higher level types.

wooorm:
Btw, @ocavue, amazing work! Thanks!

lib/character/codes.d.ts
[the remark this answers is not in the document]
ocavue: So `Code` is simulated to `Type`: a value which is not defined in the `micromark` repository can still be valid. 

Does other `Code` can also from an extra plugin like `micromark-extension-gfm`?

shared-types.d.ts
ocavue: @ChristianMurphy Do you think it's OK to just keep the way below in this PR? Since it's more simple.

```
export type Type = string
export type Code = number | null
```

lib/shared-types.d.ts
ocavue: Check if I'm right:

We have three kinds of options:
- for `lib/parse`, we have `ParseOptions`, which has one attribute:
  - extensions
- for `lib/compile/html`, we have `CompileOptions`, which has four attributes:
  - defaultLineEnding
  - allowDangerousHtml
  - allowDangerousProtocol
  - htmlExtensions
- for `lib/stream` and `lib/index`, we have `Options`, which has all five attributes above

lib/character/codes.d.ts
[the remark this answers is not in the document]
wooorm: character codes come from strings: `'🤔'.charCodeAt(n)`

shared-types.d.ts
ChristianMurphy: :thinking: that may be fine, it can be revisited if any of the extensions need a more specific type.

---

The principle I try to follow with typings is "Make invalid states unrepresentable".
Meaning not only should valid micromark calls have types be allowed, but also invalid calls to micromark should, as much as TypeScript allows, throw an error.
As long as we aren't introducing potentially invalid program states to the typings, I am all for simpler representations :+1:

lib/shared-types.d.ts
wooorm: Yes, perfect!

micromark.test.ts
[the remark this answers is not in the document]
ocavue: Just add more tests in `micromark.test.ts`. I have no experience with `.d.ts` testing, so I just add some random tests for `index.d.ts` and `stream.d.ts`. Do you think these tests are enough?

And do you think there is anything else that needs to be done for this PR?

wooorm (commented):
The code looks good to me!

I am wondering what the rules are to what is typed or not, e.g., looking at [`util/`](https://github.com/ocavue/micromark/tree/types/add-typings/lib/util), not everything has corresponding types?

ocavue:
> I am wondering what the rules are to what is typed or not.

I don't actually have rules. I just didn't add `.d.ts` for more files based on the origin PR from @ChristianMurphy.

ChristianMurphy:
> I don't actually have rules. I just didn't add .d.ts for more files based on the origin PR from @ChristianMurphy.

I was trying to work bottom up towards index.js.
I haven't/hadn't added typings for all the source files yet.
Though, since most files can/will be accessed directly having typings for most (to all) could be valuable.

wooorm:
I see three types of types:

- [x] (**priority**) types for the user interface (buffering and streaming api)
- [ ] (**somewhat important**) types for the extension interface (Constructs, codes, etc) (mostly done)
- [ ] types for random utilities. E.g., `miniflat`, or `regex-check`, which are just a tiny internal things, don’t need to be typed IMO (some done, some not)

I feel a couple of types could be removed from this PR, whereas some stuff for extensions might still be missing?

ocavue:
We can print all modules required by other repositories and check if we missed something.

A shell command like this may helps:

```shell
$ cd root_to_all_repositories
$ cat ./mdast*/**.js | grep -o "require('micromark.*" | sort | uniq
require('micromark/dist/character/codes')
require('micromark/dist/constant/constants')
require('micromark/dist/constant/has-own-property')
require('micromark/dist/constant/types')
require('micromark/dist/parse')
require('micromark/dist/postprocess')
require('micromark/dist/preprocess')
require('micromark/dist/util/normalize-identifier')
require('micromark/dist/util/safe-from-int')
```

script/copy-dict.js
```diff
@@ -0,0 +1,33 @@
...
+    var stat = fs.lstatSync(entry)
+    if (stat.isDirectory()) yield* walk(entry)
+    else if (stat.isFile()) yield entry
+  }
+}
+
+function copy(libFilePath) {
+  var distFilePath = path.join(distDir, path.relative(libDir, libFilePath))
+  fs.mkdirSync(path.dirname(distFilePath), {recursive: true}) // Need Node.js 10.12.0+
+  fs.copyFileSync(libFilePath, distFilePath)
```
ocavue, line 22: I just found out that `mdast-util-from-markdown` requires `micromark/dist/constant/constantst`, so the types should not be compiled away. I removed the logic of `@for-script: REMOVE_ALL_THING_BELOW`.

ocavue:
I ran a scan for all `micromark`  and `mdast` repositories to find out which modules are been required. Is there anything else I'm missing? @wooorm

script/copy-dict.js
wooorm: True, but that’s an open issue: https://github.com/syntax-tree/mdast-util-from-markdown/issues/4. I don’t think from-markdown or other folks should use it

ocavue: I know nothing about compilation, but shouldn't these imports be tree shaken by the compiler like Webpack? (Does Webpack have the ability to tree shake ES5 `require` imports?)

For example, `mdast-util-from-markdown` only uses two constants (`numericBaseDecimal` and `numericBaseHexadecimal`) from `micromark/dist/constant/constants`, so other constants except for these two constants should not appear in the final bundle. If my understanding is correct, two constants are too small to worry about.

wooorm: webpack indeed isn’t very good at shaking CJS, it can be better with [webpack-common-shake](https://github.com/indutny/webpack-common-shake).

In my experience, browserify is much better than webpack. Using browserify w/ tinyify shavings 1kb gzip off the bundle size, making micromark around 13.5kb instead of 14.5kb.

Still, no tool can do what we do in this project with [pre-evaluating](https://github.com/micromark/micromark#size--debug) code and removing all the constants, which is significant!

---

Regardless, I don’t see much value in typing these constants, and most of the utilities. I don’t consider them the API: they are not stable, they shouldn’t really be used. Instead, it makes it harder for me, as I don’t write typescript, to maintain this project (and other folks who may want to help out but don’t enjoy or know TS either).

I think you’re doing an amazing job here, but parts of of it feel a bit much to me? (See https://github.com/micromark/micromark/pull/25#issuecomment-705076345)

wooorm:
Re this comment: https://github.com/micromark/micromark/pull/25#issuecomment-705442402

I believe it’s not properly typed out how a sample extension would look, e.g., an extension could be this:

https://github.com/micromark/micromark-extension-math/blob/4d7956a66d2a70f9afc01aba7cdfc667cc285761/lib/syntax.js#L3-L4

Or like so:

https://github.com/micromark/micromark-extension-gfm-table/blob/210945bbe04a7d3646a6252d723a970536262bbe/syntax.js#L1-L3

So a syntax extension is an object, whose fields are keys such as `flow` or `text`, which are objects, whose fields are a serialized Code, and values are `Construct` or `Construct[]`?

script/copy-dict.js
ocavue: > I don’t consider them the API: they are not stable, they shouldn’t really be used.  

I added `@for-script: REMOVE_ALL_THING_BELOW` back. 

> parts of of it feel a bit much to me 

Do you have more detailed instructions? Like which part of `micromark` should be considered as internal and shouldn't be exported to outside?

ocavue:
> I believe it’s not properly typed out how a sample extension would look 

I will add more detailed to `SyntaxExtension` and `HtmlExtension`.

script/copy-dict.js
wooorm: I would consider `lib/*/**/*.js` private and the rest (`lib/*.js`) public. So: I see `lib/character/ascii-atext`, `lib/constant/assign`, `lib/tokenize/attention`, and the rest in subdirectories all as private.

In the case where interfaces are exposed to extensions, such as `lib/compile/html` and `lib/util/create-tokenizer`, I believe types do makes sense though.

ocavue: Just to be clear: all the files below are public, and the rest are all private. Am I getting this right?

```
lib/constructs.js
lib/index.js
lib/parse.js
lib/postprocess.js
lib/preprocess.js
lib/stream.js
lib/compile/html.js
lib/util/create-tokenizer.js 
```

------

wooorm: I’d say these are public:

```
lib/index.js
lib/parse.js
lib/postprocess.js
lib/preprocess.js
lib/stream.js
```

^-- this is what folks will deal with

Then there are things for folks making extensions. They will define SyntaxExtensions and HtmlExtensions, where they define constructs with initializers, resolvers, states, and the like, which can access stuff coming from compile/html and create-tokenizer

ocavue: However I some plugins do import more stuff from `micromark`:


https://github.com/syntax-tree/mdast-util-from-markdown/blob/a9d53edd090b71a4d50d4fababaad63d44cadaa2/index.js#L11-L12


Are these imports valid? Or these imports should be removed?

wooorm: Yes, that’s correct! But I don’t think others should do the same thing. And I think those should be removed in the future.

mdast-util-from-markdown is highly coupled to micromark. It is more a fork, or even the same thing, using hidden internals. It’s not a plugin/extension.

Some of the extensions I wrote also [use utilities](https://github.com/micromark/micromark-extension-footnote/blob/a9bbcb93fda89765851e5819ec8c8115a2108069/index.js#L3-L9) from micromark core. I don’t think that’s good though or that others should do it.

lib/shared-types.d.ts
```diff
@@ -0,0 +1,296 @@
...
+  string: Record<CodeAsKey, Construct | Construct[]>
+  text: Record<CodeAsKey, Construct | Construct[]>
+}
+
+/**
+ * https://github.com/micromark/micromark#htmlextension
+ */
+export type HtmlExtension =
+  | {enter: {[hook: string]: () => void}}
+  | {exit: {[hook: string]: () => void}}
```
ocavue, line 229: updated the types of Extensions.

lib/shared-types.d.ts
```diff
@@ -0,0 +1,296 @@
...
+
+/**
+ * https://github.com/micromark/micromark#syntaxextension
+ */
+export interface SyntaxExtension {
+  content: Record<CodeAsKey, Construct | Construct[]>
+  document: Record<CodeAsKey, Construct | Construct[]>
+  flow: Record<CodeAsKey, Construct | Construct[]>
+  string: Record<CodeAsKey, Construct | Construct[]>
+  text: Record<CodeAsKey, Construct | Construct[]>
```
wooorm, line 221: these are all optional fields

lib/shared-types.d.ts
wooorm: the keys in these objects are token types (which I think are defined somewhere in this file already)

script/copy-dict.js
ocavue: Removed all `.d.ts` for private files. Now the whole repository only has the following `.d.ts` files:

- public API
  - ./lib/index.d.ts
  - ./lib/parse.d.ts
  - ./lib/postprocess.d.ts
  - ./lib/preprocess.d.ts
  - ./lib/steam.d.ts
- Some utility type to making extensions like `SyntaxExtensions`, `HtmlExtensions`, `Construct`, `Tokenizer`, etc
  - ./lib/shared-types.d.ts
- Some constant definitions:
  - ./lib/character/codes.d.ts
  - ./lib/character/values.d.ts
  - ./lib/constant/constants.d.ts
  - ./lib/constant/types.d.ts

lib/shared-types.d.ts
ocavue: Just to make sure: Is `SyntaxExtension.content` optional?

wooorm: Yes, all fields are optional!
The keys are `document, contentInitial, flowInitial, flow, string, text`

script/copy-dict.js
wooorm: thanks for understanding, @ocavue!

To add: I think it’s wise to start small, and go from there! Maybe some of these things in utilities can be made into external tools or a public API later!

lib/shared-types.d.ts
```diff
@@ -0,0 +1,297 @@
...
+export type Event = [string, Token, Tokenizer]
+
+/**
+ * These these are transitions to update the CommonMark State Machine (CSMS)
+ */
+export interface Effects {
+  /**
+   * Enter and exit define where tokens start and end
+   */
+  enter: (type: Type) => void
```
wooorm, line 67: returns a token

lib/shared-types.d.ts
```diff
@@ -0,0 +1,297 @@
...
+
+  /**
+   * Enter and exit define where tokens start and end
+   */
+  exit: (type: Type) => void
```
wooorm, line 72: returns a token

wooorm (commented):
I think these two are my only suggestions?

I wonder what @ChristianMurphy, @fazouane-marouane think?

tmp.sh
```diff
@@ -0,0 +1,18 @@
+root="/Users/ocavue/code/github"
```
wooorm, line 1: oh btw, make sure to delete this one!

ocavue: oh my bad.

wooorm (approved):

lib/shared-types.d.ts
```diff
@@ -0,0 +1,297 @@
...
+  events: Event[]
+  interrupt: unknown
+  lazy: unknown
+  containerState: {
+    marker: number
+    type: Type
+    initialBlankLine: unknown
+    size: number
+    _closeFlow: unknown
+    furtherBlankLines: unknown
```
ChristianMurphy, line 204: do these unknowns have more concrete types?

ChristianMurphy (approved):

lib/shared-types.d.ts
wooorm: extensions can put whatever they want on there, although `_closeFlow` is a symbol to make lists work. `interrupt` and `lazy` are optional booleans. What if we allow `containerState` to be an object that allows any key/value?

ChristianMurphy (approved):
LGTM, thanks @ocavue! :bow: :+1: :tada:

PR #27, opened 2020-10-11
Migrate tests to esm

TrySound:
I migrated sources to esm locally and output size with rollup is almost the same
as with browserify.

In this diff I only migrated tests to es modules and run them with the help of
"esm" package. Later when everything will be migrated we can try to use
native node support.


wooorm:
Nice!

Can we somehow make `node test` or `node test/io/misc/bom` or so working on latest node? E.g., with `.mjs` as extensions?

TrySound:
We can make it work later when everything will be migrated to esm. `type: module` would allow us to avoid mjs extension in source code. Is this ok for you?

wooorm:
Yeah, I was indeed thinking that it wouldn’t be needed later if `type: 'module'` is set. But, as I’m not clear on when everything is ESM, it would be nice to run stuff w/o `esm` already, if it’s not needed? I don’t think there’s a downside of using `.mjs` for stuff that only runs locally in node?

TrySound:
I mean it may take a couple of days to migrate everything. A bit easier than renaming.

TrySound:
Btw native esm requires all index and extensions to be specified. So `node test` will not work. `node test/index.mjs` will.

wooorm:
> I mean it may take a couple of days to migrate everything. A bit easier than renaming.

I’m still incredibly worried about the effects of migrating to ESM on the 500 repos I maintain which are downloaded billions of times. I’m expecting the discussions to take longer than 3 days.

> node test/index.mjs will. This makes .mjs even more annoying and harder to get back .js extension.

Would this not also affect `.js`? And it’s possible to autocomplete paths, which makes it simpler to call than having to add `-r esm`. Finally, it’s standard Node instead of a polyfills. I don’t see the downside 🤔

ChristianMurphy:
cross linking, this is related to https://github.com/unifiedjs/rfcs/pull/4#issuecomment-679233818

ChristianMurphy:
Thanks @TrySound!
Definitely interested in bringing native ESM support into micromark and the unified ecosystem. :+1: 

It's a big move, a lot of the npm ecosystem seems fragmented, as bits and pieces migrate over, unified is also a pretty large ecosystem, I'm interested in how this works on a broader scale across the 300+ packages in unified that may also need to migrate.
Some of my question may be a bit broader than just this PR as we think this through.

> and output size with rollup is almost the same as with browserify.

I'm not sure I follow this comment, if this affects only the tests, and source is not modified how would it impact the bundle size?

> In this diff I only migrated tests to es modules and run them with the help of "esm" package.

Help me understand how this helps move the migration to ESM forward.
I'm not as familiar with the migrate tests first approach, it seems, perhaps naively, that we'd be testing in a way most adopters would not be using the library, while not actually changing anything about the library? :confused: 

> Later when everything will be migrated we can try to use native node support.

I'm cautious a lot of project are still on CJS, this sounds like it would drop CJS support entirely.
Would tools like: https://github.com/developit/microbundle, https://github.com/mikeal/limbo, or https://github.com/pikapkg/snowpack help here?
Ideally there'd be some overlap time between CJS and ESM to allow folks time to migrate, rather than an immediate cut off of CJS support.

TrySound:
> I'm not sure I follow this comment, if this affects only the tests, and source is not modified how would it impact the bundle size?

I migrated sources to esm **locally**

> I'm cautious a lot of project are still on CJS, this sounds like it would drop CJS support entirely.

I meant only everything in this particular repository. I'm not gonna give up on commonjs users. Both cjs and esm will be distributed.

> Would tools like: https://github.com/developit/microbundle, https://github.com/mikeal/limbo, or https://github.com/pikapkg/snowpack help here?

I'm gonna use rollup which is a base for all these tools.

TrySound:
Alright, I migrated tests to mjs extension. Though json modules do not work yet in esm so I added proxy .cjs modules to fix this. Fuzz tests does not work with native modules, I used "esm" to make it work.

Also looks like native esm does not work in node 10. Can I drop it?

.travis.yml
```diff
@@ -1,5 +1,4 @@
 language: node_js
 node_js:
-  - lts/dubnium
```
wooorm, line 3: I was under the impression that dubnium does support `.mjs`? It didn’t work for you?

Might it be better to set another lower barrier? Erbium?

test/io/misc/character-entities.cjs
```diff
@@ -0,0 +1 @@
+module.exports = require('character-entities')
```
wooorm, line 1: I do think it’s a bug in Node that it can’t require clear CJS (as the package has no `type:`).
They do seem to have a [JSON loader](https://github.com/nodejs/node/blob/2d83e743d934738a3ad26ed587eb7cfd6ca46081/lib/internal/modules/esm/translators.js#L273), so I’m wondering why it’s not used here.

package.json
```diff
@@ -96,10 +97,10 @@
...
-    "test-fuzz": "timeout 15m jsfuzz test/fuzz.js test/fixtures --rss-limit-mb 300 --timeout 300",
-    "test-api": "node test",
-    "test-coverage": "nyc --reporter lcov tape test/index.js",
+    "pretest-fuzz": "patch-package --patch-dir script/patches; node script/generate-fixtures.mjs",
+    "test-fuzz": "timeout 15m NODE_OPTIONS='-r esm' jsfuzz test/fuzz.js test/fixtures --rss-limit-mb 300 --timeout 300",
+    "test-api": "node test/index.mjs",
+    "test-coverage": "nyc --reporter lcov node test/index.mjs",
     "test-types": "dtslint .",
     "test": "npm run generate && npm run format && npm run test-coverage && npm run test-types"
   },
```
wooorm, line 106: I did not have luck setting:

```
    "extensions": [
      ".cjs",
      ".mjs"
    ],
```

in `xo` below, but it did work passing them directly to `xo` in the `format` script above.

The last section on the xo overrides deal with the tests. I think it should be:

```
      {
        "files": [
          "test/**/*.{js,mjs,cjs}"
        ],
        "rules": {
          "import/extensions": "off",
          "import/no-unassigned-import": "off",
          "import/no-useless-path-segments": "off",
          "unicorn/import-index": "off"
        }
      }
```

package.json
```diff
@@ -96,10 +97,10 @@
     "generate-size": "browserify . -p tinyify -s micromark -o micromark.min.js; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
     "generate": "npm run generate-constant-typings && npm run generate-expressions && npm run generate-dist && npm run generate-size",
     "format": "remark . -qfo && prettier . -w --loglevel warn && xo --fix",
-    "pretest-fuzz": "patch-package --patch-dir script/patches; node script/generate-fixtures",
-    "test-fuzz": "timeout 15m jsfuzz test/fuzz.js test/fixtures --rss-limit-mb 300 --timeout 300",
-    "test-api": "node test",
-    "test-coverage": "nyc --reporter lcov tape test/index.js",
+    "pretest-fuzz": "patch-package --patch-dir script/patches; node script/generate-fixtures.mjs",
+    "test-fuzz": "timeout 15m NODE_OPTIONS='-r esm' jsfuzz test/fuzz.js test/fixtures --rss-limit-mb 300 --timeout 300",
```
wooorm, line 101:
```suggestion
    "test-fuzz": "NODE_OPTIONS='-r esm' timeout 15m jsfuzz test/fuzz.js test/fixtures --rss-limit-mb 300 --timeout 300",
```

.travis.yml
wooorm: If possible, I would prefer to still test on dubnium

test/io/misc/character-entities.cjs
TrySound: https://nodejs.org/api/esm.html#esm_experimental_json_modules

package.json
TrySound: It's probably the latest eslint behaviour

.travis.yml
TrySound: Erbium is passed.

dubnium works with `--experimental-modules` flag though .cjs extension is not supported when importing from .mjs.
Anyway this should not affect node 10 users if you stick with es5. We just should not add .cjs proxies.

test/io/misc/character-entities.cjs
wooorm: I think this is different: you can import CJS from MJS. CJS should then be able to work as it used to, no?
Node knows this module is CJS, and it can require JSON files in CJS, but it seems to break on json-only modules

.travis.yml
wooorm: > Erbium is passed.

What do you mean with this?

wooorm: oh whoops, sorry, missed that erbium is below this line.

TrySound: Erbium tests are passed :)

test/io/misc/character-entities.cjs
TrySound: I hope node team will backport stabilized json modules to node 12

wooorm: Is this something that should be raised with them perhaps?

wooorm (approved):
I think this looks great, except for the wart that is JSON in CJS from ESM, which is out of our control.

Other stuff was raised here too for about moving the rest of the code. I want to move discussing that to another issue or PR once it’s there.

ChristianMurphy:
> Also looks like native esm does not work in node 10. Can I drop it?

@wooorm can chime in here, but IIRC, we've supported release lines until EOL in the past.
Node 10 doesn't reach EOL until mid -021 https://github.com/nodejs/Release#release-schedule

wooorm:
@ChristianMurphy that was already resolved with https://github.com/micromark/micromark/pull/27/commits/2670fe53e1b0574ae7c6592cb78b014e4d8a7a60 I believe!

ChristianMurphy:
ah, nvm :sweat_smile:

wooorm:
@ChristianMurphy :+1: Any blocking thoughts on this or good to go?

ChristianMurphy (approved):

wooorm:
Thanks Bogdan!

test/io/misc/character-entities.cjs
MylesBorins: We went into it in depth on twitter but TL;DR the current experimental modules will not likely stabilize in their current form as JSON modules are being standardized by TC39 right now using import assertions due to security needs of browsers

https://github.com/tc39/proposal-json-modules

PR #28, opened 2020-10-12
Replace `browserify` with `rollup`

TrySound:
Browserify: 13602 b
Rollup: 13585 b

Rollup is the best bundler to deal with es modules. In this diff I setup
it before migrating sources to esm. Achieved size is almost the same.

The config will be reused to produce cjs/esm entry points.


package.json
```diff
@@ -64,9 +64,11 @@
   "devDependencies": {
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
+    "@rollup/plugin-babel": "^5.2.1",
+    "@rollup/plugin-commonjs": "^15.1.0",
+    "@rollup/plugin-node-resolve": "^9.0.0",
```
wooorm, line 69:
```suggestion
    "@rollup/plugin-babel": "^5.0.0",
    "@rollup/plugin-commonjs": "^15.0.0",
    "@rollup/plugin-node-resolve": "^9.0.0",
```

Assuming we’re not using any fancy new features, zero-filling will reduce some noice in the future

wooorm (commented):
Any idea where the couple of bytes reduction is coming from? Is it just because this version doesn‘t build as a standalone-UMD bundle?

wooorm:
Looks like the bundle is output to stdout(4) on CI btw

TrySound:
Actually I saw the same with browserify. Looks like something else prints code. Will check.

TrySound:
The reduction comes from stripped debug I guess. It was not stripped with browserify in configuration.

wooorm:
🤔 I didn't see that with browserify. It had a -o flag which wrote to a file I think?

wooorm:
Debug is removed in the dist files already w babel, so that can't be it!

TrySound:
lib was the input to browserify, not dist. So the size is grown a bit because of commonjs but we will achieve more optimisation after migrating to esm.

wooorm:
Are you 100%? I checked that before. '.' Resolves to dest. You're saying this removes 20 bytes, but debug is much bigger than that so I have a hard time seeing it 🤔 tinyify is amazing, I can only understand the 20 byte shave coming from dropping the UMD bundle

TrySound:
Ah, I see. You are right. I though main points to lib/index.js. Well, then I guess rollup just produces more optimal code then.

TrySound:
Ah, and now it's in esm format, not cjs or umd.

TrySound:
tinify is a set of optimisations around browserify + two passes of uglify. Btw I copy pasted it and passed to terser. Rollup does all optimisations out of the box.

TrySound:
Here's per format sizes
Browserify: 13602 b
Rollup esm: 13585 b
Rollup cjs: 13594 b
Rollup umd: 13685 b

wooorm:
Hi thanks for your patience! I wanted to take a break from this repo after having lived in here for several months 😅

> Here's per format sizes
> Browserify: 13602 b
> Rollup esm: 13585 b [**100b** off from rollup umd, **17b** off from browserify umd]
> Rollup cjs: 13594 b [**91b** off from rollup umd, **8b** off from browserify umd]
> Rollup umd: 13685 b [**83b** bigger than browserify umd]

From these numbers, I conclude:
a) browerify does not output CJS/ESM or this isn’t checked.
b) rollup performs worse than browserify on comparable targets
c) the saving come from checking a different target: CJS or ESM over UMD.

So, what is the advantage of using rollup over browserify? The file from `generate-size` is not published: the numbers are used to check how big micromark would be in a browser. Switching to a different bundler/minifier will yields slightly different results, but that is unrelated to micromark.

TrySound:
Afaik browserify cannot handle es modules without transpilation. Rollup supports only es modules out of the box. Migrating to esm will allow to scope hoist utilities, constants etc and make them mangleable. Browserify will probably not be able to show such reduction.

Ps: Im on vacation for the next two weeks. Cannot rebase.

wooorm:
> Afaik browserify cannot handle es modules without transpilation. Rollup supports only es modules out of the box. 

 Both true I believe. 
> Migrating to esm will allow to scope hoist utilities, constants etc and make them mangleable. Browserify will probably not be able to show such reduction. 

I strongly doubt that ESM improves these things. The code is already written with this in mind, and common-shakeify can do it too. ESM is an improvement over UMD—it can shave those 17 bytes off in this bundle we’re checking. But that is not a cost anyone actually has: if you bundle code, you’re not getting those bytes because you’re not getting an UMD bundle. In fact, adding ESM would add bytes for every file for every user: https://github.com/unifiedjs/rfcs/pull/4#issuecomment-679919164? I do see ESM as the way forward. I personally don’t see it as better in any way though. It’s just the blessed way.

If I am correct, and you’re interested in working on using ESM for micromark and compiling it with Rollup to CJS, then we’re getting another rollup config, correct? Would in then make sense to first do that while keeping browserify/tinyify (and maybe even keep on doing that in the future, or using both, so that we keep on comparing the same bundle that folks will typically get?)

> Ps: Im on vacation for the next two weeks. Cannot rebase.

Have fun, take care, and please don‘t respond yet unless you want to!

TrySound:
Gotcha. The problem was here

https://github.com/micromark/micromark/blob/main/lib/tokenize/block-quote.js#L50
https://github.com/micromark/micromark/blob/main/lib/tokenize/list.js#L155

Rollup wraps such dynamic usages with commonjs runtime which increased bundle size.

Now rollup umd bundle is 13566 (against 13602 with browserify)

TrySound:
Hi @wooorm Is it good to go?

package.json
```diff
@@ -85,17 +87,18 @@
...
   "scripts": {
     "generate-constant-typings": "node script/generate-constant-typings",
     "generate-expressions": "node script/generate-expressions",
-    "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier dist --loglevel error --write",
-    "generate-size": "browserify . -p tinyify -s micromark -o micromark.min.js; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
+    "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier --loglevel error --write dist",
+    "generate-size": "rollup -c; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
     "generate": "npm run generate-constant-typings && npm run generate-expressions && npm run generate-dist && npm run generate-size",
-    "format": "remark . -qfo && prettier . -w --loglevel warn && xo --fix",
+    "format": "remark . -qfo && prettier --write . --loglevel warn && xo --fix",
```
wooorm, line 101:
```suggestion
    "format": "remark . -qfo && prettier . -w --loglevel warn && xo --fix",
```

lib/tokenize/block-quote.js
```diff
@@ -1,5 +1,8 @@
-exports.tokenize = tokenizeBlockQuoteStart
-exports.continuation = {tokenize: tokenizeBlockQuoteContinuation}
+var tokenize = tokenizeBlockQuoteStart
+var continuation = {tokenize: tokenizeBlockQuoteContinuation}
```
wooorm, line 2: Would it work to have a `var blockQuote = {tokenize: tokenize, continuation: continuation, exit: exit}` here, export that, and use it in the continuation?

Or only this at the top:

```
var blockQuote = {
  tokenize: tokenizeBlockQuoteStart,
  continuation: tokenizeBlockQuoteContinuation,
  exit: exit
}
```

...and exporting that, and using that in continuation?

lib/tokenize/list.js
```diff
@@ -1,6 +1,10 @@
-exports.tokenize = tokenizeListStart
-exports.continuation = {tokenize: tokenizeListContinuation}
-exports.exit = tokenizeListEnd
+var tokenize = tokenizeListStart
```
wooorm, line 1: same q as for block quote

rollup.config.js
```diff
@@ -0,0 +1,47 @@
...
+      }),
+      terser({
+        output: {
+          // eslint-disable-next-line camelcase
+          ascii_only: true
+        },
+        mangle: {
+          safari10: true
+        }
+      })
```
wooorm, line 30: Do you see differences for using `terser` twice? tinyify is running it on separate files, and on the whole, but this does the same twice. Maybe this isn’t noticable?

wooorm (commented):
Hi! Thanks for your patience. Couple Qs and I’ll try this out locally now!

wooorm:
Also: running `npm test` on the code here reformats a couple of comments/style things

package.json
```diff
@@ -85,17 +87,18 @@
...
-    "tinyify": "^3.0.0",
     "xo": "^0.34.0"
   },
   "scripts": {
     "generate-constant-typings": "node script/generate-constant-typings",
     "generate-expressions": "node script/generate-expressions",
-    "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier dist --loglevel error --write",
-    "generate-size": "browserify . -p tinyify -s micromark -o micromark.min.js; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
+    "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier --loglevel error --write dist",
+    "generate-size": "rollup -c; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
```
wooorm, line 99:
```suggestion
    "generate-size": "rollup -c --silent; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
```

The default output by rollup is a bit more verbose than I’d like. Silent sounds a bit harsh, maybe there’s something in between, or --silent is good enough?

package.json
```diff
@@ -85,17 +87,18 @@
...
     "tape": "^5.0.0",
-    "tinyify": "^3.0.0",
     "xo": "^0.34.0"
   },
   "scripts": {
     "generate-constant-typings": "node script/generate-constant-typings",
     "generate-expressions": "node script/generate-expressions",
-    "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier dist --loglevel error --write",
-    "generate-size": "browserify . -p tinyify -s micromark -o micromark.min.js; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
+    "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier --loglevel error --write dist",
```
wooorm, line 98:
```suggestion
    "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier dist --loglevel error --write",
```

Unrelated change

package.json
TrySound: This logs all output instead of writing.

TrySound:
```
$ remark . -qfo && prettier . --write --loglevel warn && xo --fix
Error: Failed to load plugin 'prettier' declared in 'CLIOptions': Cannot find module 'eslint-plugin-prettier'
```

rollup.config.js
TrySound: 13425 vs 13589. But yeah, single pass is more realistic for user build.

wooorm:
`rm -rf node_modules && npm i`?

TrySound:
It works. Thanks. Btw lockfile allows to avoid this.

lib/tokenize/block-quote.js
TrySound: So you suggest to make every tokenize a default export?

wooorm:
> lockfiles

I have only heard of this error for folks using yarn. Can you confirm whether that was the case?

---

Comparing your branch to main, I’m now getting:

```
rollup:
13.4 kB
13436

browserify / tinyify:
13.5 kB
13479
```

Rollup/terser removes 43 bytes! Nice, but pretty small. Again: ESM is the future, but that is soooo small. ESM isn’t better than CJS at tree shaking 🙄

TrySound:
> I have only heard of this error for folks using yarn. Can you confirm whether that was the case?

Ah, actually you are right. I'm tied to yarn lockfiles 😄 

> Rollup/terser removes 43 bytes! Nice, but pretty small.

It's 13386 with simplified blockquote and list exports. And remember, the code is still commonjs. Rollup may do less assumptions about safe commonjs transformations.

wooorm:
Yeah, you’re right! Maybe that’ll help!

package.json
```diff
@@ -85,17 +87,18 @@
...
   },
   "scripts": {
     "generate-constant-typings": "node script/generate-constant-typings",
     "generate-expressions": "node script/generate-expressions",
     "generate-dist": "babel lib --out-dir dist --quiet --retain-lines --plugins babel-plugin-unassert,./script/babel-transform-undebug.js,./script/babel-transform-constants.js; node script/copy-dict.js; prettier dist --loglevel error --write",
-    "generate-size": "browserify . -p tinyify -s micromark -o micromark.min.js; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
+    "generate-size": "rollup -c --silent; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
     "generate": "npm run generate-constant-typings && npm run generate-expressions && npm run generate-dist && npm run generate-size",
-    "format": "remark . -qfo && prettier . -w --loglevel warn && xo --fix",
+    "format": "remark . -qfo && prettier . --write --loglevel warn && xo --fix",
```
wooorm, line 101:
```suggestion
    "format": "remark . -qfo && prettier . -w --loglevel warn && xo --fix",
```

haha, you really like the long form 😅 If you feel strongly, I’m fine with it. Otherwise, you can apply this suggestion

rollup.config.js
wooorm: ohhh, that’s significant! Weird! Well, let’s leave it in then

package.json
TrySound: Ah, short form works with reinstalled node_modules. Probably was a bug in old prettier. But yeah, let's use long version for consistency :)

wooorm: Uhm, the consistency is that short flags are used where possible, and it was that way before, hence why I’m suggesting to go back to how it was 😅

Anyway, feel free to commit this suggestion, or not. I think the rest looks good. Anything else?

TrySound: In other script --write was used

wooorm (approved):

wooorm:
Awesome, thanks @TrySound!

PR #34, opened 2020-11-10
types: skip lib check

ChristianMurphy:
we test all the types we use in https://github.com/micromark/micromark/blob/main/micromark.test.ts, we may not need to check the types we don't use.


fazouane-marouane (approved):

wooorm (approved):

wooorm:
Thanks @ChristianMurphy!

@TrySound Your PR #28 should now be green once rebased

PR #35, opened 2020-11-19
Add es modules support

TrySound:
In this diff I provided es modules and commonjs builds with rollup.
File structure is preserved to allow ecosystem use micromark internals.
(hope it's temporary solution).

Also provided "exports" fields with "/buffer" and "/stream" imports
support. `"./": "./"` keeps all internals available (usually "exports" is
used to mask available imports).

All tests point to mjs versions of code.

Added es modules support for babel transforms and covered them with
tests.

I removed asciiAlphaCaseDifference in constants because rollup
treeshakes it out from dist.


test/babel-transform-undebug.mjs
```diff
@@ -0,0 +1,46 @@
+import path from 'path'
+import test from 'tape'
+import babel from '@babel/core'
+import transformConstants from '../script/babel-transform-undebug.mjs'
+
+function transformCode(code) {
+  const result = babel.transformSync(code, {
```
wooorm, line 7: maybe have a `configFile: false` here? https://babeljs.io/docs/en/options#configfile

Recently working on projects with babel, and it can pick up babel configs, which in this case it shouldn’t I believe?

test/babel-transform-undebug.mjs
```diff
@@ -0,0 +1,46 @@
...
+    plugins: [transformConstants],
+    filename: path.resolve('test/babel-transform-undebug.mjs')
```
wooorm, line 9:
```suggestion
    filename: path.resolve(path.join('test', 'babel-transform-undebug.mjs'))
```

windows

test/babel-transform-undebug.mjs
```diff
@@ -0,0 +1,46 @@
...
+      t.plan(1)
+      var input = `
+      import createDebug from 'debug'
+      var debug = createDebug('micromark')
+      debug('log')
+      console.log(\'log\')
+      `
+      var output = 'console.log(\'log\');'
+      var result = transformCode(input)
+      t.equal(result, output)
```
wooorm, line 41: I’d prefer the names `actual` for `result` and `expected` for `output`, to closer reflect typical testing names

test/babel-transform-constants.mjs
```diff
@@ -0,0 +1,145 @@
+import path from 'path'
+import test from 'tape'
+import babel from '@babel/core'
+import transformConstants from '../script/babel-transform-constants.mjs'
```
wooorm, line 4: same suggestions as for `debug` tests

wooorm: and thanks for adding tests for these!!! ✨

script/generate-expressions.mjs
```diff
@@ -0,0 +1,48 @@
+import fs from 'fs'
+import path from 'path'
+import regenerate from 'regenerate'
+import codes from '../lib/character/codes.js'
```
wooorm, line 4: Could you add `codes` below all `@unicode` imports? I guess I missed that before. But I prefer local files last.

script/babel-transform-constants.mjs
```diff
@@ -1,27 +1,62 @@
...
 
 var supported = [
-  require.resolve('../lib/character/codes'),
-  require.resolve('../lib/character/values'),
-  require.resolve('../lib/constant/constants'),
-  require.resolve('../lib/constant/types')
+  'micromark/lib/character/codes.js',
+  'micromark/lib/character/values.js',
+  'micromark/lib/constant/constants.js',
+  'micromark/lib/constant/types.js'
```
wooorm, line 20: The reason for this handling is because of: https://github.com/syntax-tree/mdast-util-from-markdown/issues/4, which landed in https://github.com/micromark/micromark/commit/bf53bf93074068376240dc692267f1082325bfd4.

I’m assuming your change works fine in this project (although: windows?), but I’m wondering if it’d also work on other projects using this script and having the files in `node_modules/micromark/...`?

script/babel-transform-constants.mjs
```diff
@@ -1,27 +1,62 @@
...
+    var local
+    var position
+
+    actual = resolveFrom(dirname, p.node.source.value)
+    p.node.specifiers.forEach((specifier) => {
+      if (specifier.type === 'ImportDefaultSpecifier') {
+        id = specifier.local.name
+      } else {
+        throw Error(
+          'Unknown specifier "' + specifier.type + '" in "' + p.toString() + '"'
```
wooorm, line 47:
```suggestion
          'Unknown specifier "' + specifier.type + '" in "' + String(p) + '"'
```

doubt it would happen but this prevents `undefined.toString()` throwing weird errors

rollup.config.mjs
```diff
@@ -0,0 +1,83 @@
...
+  })
+}
+
+if (process.env.BUILD === 'size') {
+  configs.push({
+    input: './lib/index.js',
+    output: {
+      file: './micromark.min.js',
+      format: 'umd',
+      name: 'Micromark',
```
wooorm, line 56:
```suggestion
      name: 'micromark',
```

I may have missed this before, but it’s not a class—the examples use `micromark`, so it’d be good to match that casing

script/babel-transform-constants.mjs
```diff
@@ -1,11 +1,17 @@
...
-import types from '../lib/constant/types.js'
+
+// TODO replace with regular imports after migrating lib to es modules
+var requireUtil = module.createRequireFromPath(
+  path.join(process.cwd(), './script/babel-transform-constants.mjs')
+)
+var codes = requireUtil('../lib/character/codes.js')
+var values = requireUtil('../lib/character/values.js')
+var constants = requireUtil('../lib/constant/constants.js')
+var types = requireUtil('../lib/constant/types.js')
```
TrySound, line 14: This hack prevents rollup from bundling commonjs in node < 13

package.json
```diff
@@ -93,14 +117,14 @@
...
-    "generate-size": "rollup -c --silent; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
+    "generate-constant-typings": "node script/generate-constant-typings.mjs",
+    "generate-expressions": "node script/generate-expressions.mjs",
+    "generate-dist": "BUILD=dist rollup -c --silent; node script/copy-dict.js; prettier dist --loglevel error --w",
+    "generate-size": "BUILD=size rollup -c --silent; gzip-size micromark.min.js; gzip-size --raw micromark.min.js",
     "generate": "npm run generate-constant-typings && npm run generate-expressions && npm run generate-dist && npm run generate-size",
     "format": "remark . -qfo && prettier . -w --loglevel warn && xo --fix",
     "pretest-fuzz": "patch-package --patch-dir script/patches; node script/generate-fixtures.mjs",
-    "test-fuzz": "NODE_OPTIONS=\"-r esm\" timeout 15m jsfuzz test/fuzz.js test/fixtures",
+    "test-fuzz": "NODE_OPTIONS=\"-r esm\" jsfuzz test/fuzz.js test/fixtures",
```
wooorm, line 127: could you revert this change? I’ve documented how it can be installed (and this script used) on unix now: https://github.com/micromark/micromark#test

package.json
```diff
@@ -44,25 +44,49 @@
     "Mudit Ameta <zeusdeux@gmail.com> (https://mudit.xyz)",
     "Christian Murphy <christian.murphy.42@gmail.com>",
     "Marouane Fazouane <fazouanem3@gmail.com>",
-    "Merlijn Vos <merlijn@soverin.net>"
+    "Merlijn Vos <merlijn@soverin.net>",
+    "Bogdan Chadkin <trysound@yandex.ru>"
```
wooorm, line 48: ✨✨✨

test/babel-transform-undebug.mjs
TrySound: AFAIK resolve normalizes slashes

script/generate-expressions.mjs
TrySound: Sure, me too. Usually rely on eslint-plugin-import though hate it dependency tree

script/babel-transform-constants.mjs
TrySound: Btw, I always thought windows just works with direct slashes. Backslash is just default.

TrySound: I don't get from the commit. Where micromark is in mdast-util-from-markdown?

wooorm:
Could you expand on what `"./": "./"` does? I don’t understand 🤔

script/babel-transform-constants.mjs
TrySound: These strings are not used for importing/requiring/reading so safe to use in windows.

wooorm: Node handles `/` in requires, but not in path related stuff.

For the commits/prs I references: it’s indeed rather complex. `mdast-util-from-markdown` is using the micromark parser, but adding a different “compiler”: it doesn’t go to HTML but goes to an AST. This is not documented (yet) — most folks won’t do it, but it is powerful and intended to be used. To make writing `from-markdown` easier, I’m also using constants, types, etc there. But: to make it fast, I also want to compile them away: https://github.com/syntax-tree/mdast-util-from-markdown/commit/e59424359047d9ee47b83f090665f93d2c23349e#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R5-R8

TrySound:
With new "exports" field we can easily hide internals from importing. For example this will prevent from importing "micromark/buffer.js" or "micromark/dist/buffer" etc.
```
{
  exports: {
    "./buffer": {
      "require": "./buffer.js"
      "import": "./buffer.mjs"
  }
}
```

`"./": "./"` keep all modules available as before. I saw you used internals in mdast-util-from-markdown package.

test/babel-transform-undebug.mjs
wooorm: I thought it didn’t (I responded somewhere else earlier). I’ll add tests on Windows in Actions later, or you could if you want to check

script/babel-transform-constants.mjs
TrySound: I don't understand how this works. How micromark happens in `./micromark`?
```
babel lib/ --out-dir dist/ --quiet --retain-lines --plugins ./micromark/script/babel-transform-constants.js; prettier dist/ --loglevel error --write
```

TrySound:
Fixed tests

script/babel-transform-constants.mjs
wooorm: it’s cloned in from-markdown: https://github.com/syntax-tree/mdast-util-from-markdown/blob/264b0c015db1835470cb42ed66b0a1564e837c4a/package.json#L63

A bit hacky, totally 😅

TrySound: Hm... I would protect it with tag version

wooorm: Sorry, I think I’ve explained this rather rather badly.

I don’t think it needs to be protected. I think the problem is: two projects use `babel-transform-constants.mjs`:

- micromark itself, in which case it should strip:
  - `micromark/lib/character/codes.js`
  - `micromark/lib/character/values.js`
  - ...
- and `mdast-util-from-markdown`, in which case it should strip:
  - `mdast-util-from-markdown/node_modules/micromark/lib/character/codes.js`
  - `mdast-util-from-markdown/node_modules/micromark/lib/character/values.js`
  - ...

Hence, the code here used `require.resolve`.

The fact that `mdast-util-from-markdown` uses git to get the babel scripts, and whether it’s git clone or a submodule, shouldn’t matter. A different approach to get it in from-markdown would be to `curl https://raw.githubusercontent.com/micromark/micromark/main/script/babel-transform-constants.js > script/babel-transform-constants.js`.

TrySound: Maybe easier to keep both project in the same repository? Babel script problem will go away. And it would be easier to  upgrade from-markdown on micromark changes.

wooorm: That is an unrelated problem. If they were in a monorepo, then lerna or whatever would put `repo/packages/mdast-util-from-markdown/node_modules/micromark` in from-markdown.

I don’t think you need to change anything in from-markdown. The problem is in your change here, where you removed `require.resolve`. What is the reason `require.resolve` cannot be used?

TrySound: Es modules does not have require

TrySound: Btw, yarn monorepo links all packages. So `repo/packages/mdast-util-from-markdown/node_modules/micromark` === `repo/packages/micromark`. NPM 7 probably does the same. Though there would be easier to setup script in the root of the project. And easier to check if it works. And from-markdown build step would not be immediately broken after merging this PR.

wooorm: Ah, right, ok let me debug this some more.

One thought: could the path from `import.meta` be used with `resolve-from`?

wooorm: There is a function coming in Node, but on latest Node it’s still under a flag: https://nodejs.org/api/esm.html#esm_no_require_resolve

wooorm: Quick Q if you have the time: Can babel support ESM plugins? `babel ... --plugins ./micromark/script/babel-transform-constants.mjs` crashes

TrySound: Not yet. Babel 8 will.

TrySound: Hm, I missed import.meta.url is supported in node 10. Will try to make it work.

TrySound: One problem is createRequireFromPath does not accept import.meta.url. createRequire which accepts is supported only since node 12.

TrySound: Oh, this solves the problem
https://nodejs.org/docs/latest-v10.x/api/url.html#url_url_fileurltopath_url

wooorm: Ohh yeah, I~’m now looking at your new rollup config and was indeed thinking `createRequire` might work

TrySound: It works. I copy pasted rollup config to from-markdown and all works with the last commit.

TrySound: Ah, shit, does not work in dubnium because of rollup.
Anyway this will be fixed when all lib will migrate to modules.
Btw I already migrated everything locally. Is it ok to merge as is?

TrySound: I reverted latest commit

wooorm: As in, from-markdown fails on dubnium?

TrySound: Both fails yet with import.meta.url so I reverted to hardcoded path which does not work in from-markdown. I cannot fix it in this PR but will do this in the next with migrating lib to es modules.

wooorm (approved):
OK. All looks good except that I’m not happy with the broken script on dubnium, but we can solve it later.

TrySound:
Now it works here. Only from-markdown is broken.

wooorm:
✨

Let me try and add windows to the GH Actions tests

PR #36, opened 2020-11-20
Migrate all source code to es modules

TrySound:
Final touch. For some files git does not show diff sadly because their
content is signinficantly changed.


wooorm:
Alright, main works on Windows ✨

lib/character/codes.d.ts
```diff
@@ -5,60 +5,28 @@ export type Code = null | number
 // @for-script: REMOVE_ALL_THING_BELOW
 
 export interface Codes {
```
fazouane-marouane, line 7: Is there a way to not reorder this ?

fazouane-marouane:
Hi 👋
Is it necessary to rename files into `.mjs` since we bundle the code using rollup ?

fazouane-marouane:
Also I wonder if having this change is enough as per https://nodejs.org/api/packages.html#packages_type
```json
// package.json
{
  "type": "module"
}
```

TrySound:
For sure I would prefer to not rename modules though it is the only way to support es modules in node 10.

TrySound:
We came to such decision here https://github.com/micromark/micromark/pull/27

lib/character/codes.d.ts
TrySound: Looks like named exports are read in alphabetical order

wooorm: Hmm. While I get alphabetical, it makes the file rather unreadable. Would be preferable if we could find something for it

wooorm:
@fazouane-marouane Reason for .mjs is that in most cases rollup is *not* used. Actual untranspiled ESM is running in Node 10 and later!

lib/character/codes.d.ts
wooorm: Probably by adding a sort here? https://github.com/micromark/micromark/blob/b73fa448123f4c4b261318caa434f39c1195decf/script/generate-constant-typings.mjs#L11

lib/character/values.d.ts
```diff
@@ -1,25 +1,16 @@
 // This module is generated by `script/`.
 
 export type Value =
```
wooorm, line 3: Same as the other generated type file: sorting might help?

rollup.config.mjs
```diff
@@ -45,23 +47,22 @@ if (process.env.BUILD === 'dist') {
     },
     external: (id) => !id.startsWith('.') && !path.isAbsolute(id),
     plugins: [
-      nodeResolve({browser: true}),
       babel({
         babelHelpers: 'bundled',
```
wooorm, line 51: btw: I hope no helpers are injected by babel? 🤔  Should we maybe use `external` to ensure things crash if they accidentally are added? https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers

test/babel-transform-constants.mjs
```diff
@@ -104,13 +104,19 @@ test('babel-transform-constants', function (t) {
...
         )
       },
       /Unknown specifier/,
       'should throw on missing fields (specifier)'
     )
 
+    t.equal(
+      transform('import { EventEmitter } from "events"\nconsole.log(0)'),
+      'import { EventEmitter } from "events";\nconsole.log(0);',
+      'should support types'
```
wooorm, line 117:
```suggestion
      'should not strip non-constants'
```

Is this message clearer for what you]re checking? Something else?

lib/character/codes.d.ts
TrySound: Fixed by using default export. In object keys order is preserved.

TrySound:
All is done

test/babel-transform-constants.mjs
```diff
@@ -15,31 +15,31 @@ test('babel-transform-constants', function (t) {
   t.test('cjs', function (t) {
     t.equal(
       transform(
-        'var codes = require("../lib/character/codes.js")\nconsole.log(codes.carriageReturn)'
+        'var codes = require("../lib/character/codes.mjs")\nconsole.log(codes.carriageReturn)'
```
wooorm, line 18: One q: as this is the `cjs` test and a require, I guess these should remain `.js`?

Otherwise: all looks good.

I’ll leave this open for a couple days so other folks can review.

After that I’ll do a bit of refactoring.

And then, ready for a minor release I guess? Anything else?

TrySound: Just to pass tests. I don't think require logic is necessary here now.

I think all is done.

wooorm: ah right, because there is no `.js` in `lib/`. Unfortunate.
There is one in `dist/` though, but that would mean changing the constants plugin, which can be left for another time

wooorm (approved):
all looks good.

I’ll leave this open for a couple days so other folks can review.

After that I’ll do a bit of refactoring.

And then, ready for a minor release I guess? Anything else?

rollup.config.mjs
```diff
@@ -45,23 +47,23 @@ if (process.env.BUILD === 'dist') {
     },
     external: (id) => !id.startsWith('.') && !path.isAbsolute(id),
     plugins: [
-      nodeResolve({browser: true}),
       babel({
-        babelHelpers: 'bundled',
+        babelHelpers: 'external',
+        skipPreflightCheck: true,
```
ChristianMurphy, line 52: were preflight checks causing an issue? https://github.com/rollup/plugins/tree/master/packages/babel#skippreflightcheck
They seem harmless, if not beneficial.

TrySound: Which issue? No issue. Preflight check just does not work with external helpers which are usually incorrect to use.

ChristianMurphy: And remind me why the helpers need to be external :thinking:

TrySound: https://github.com/micromark/micromark/pull/36#discussion_r527732630

wooorm: Might be an XY problem: I wanted to ensure we don’t get helpers in there. Maybe there’s a better way to ensure that!

TrySound: Well, it may not always fail. But I can add plugin which will fail if babel/runtime import occurs.

TrySound: Or just watch carefully for plugin list. Helpers are appeared only from builtin plugins usually.

PR #38, opened 2020-11-23
Move "debug" package to devDependencies

fazouane-marouane:
Since the "debug" package is compiled away, we may as well make it a devDependency. (Related to https://github.com/micromark/micromark/issues/37 )

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "./": "./"
   },
   "dependencies": {
-    "debug": "^4.0.0",
     "parse-entities": "^2.0.0"
   },
   "devDependencies": {
@@ -98,6 +97,7 @@
     "commonmark.json": "^0.29.0",
     "concat-stream": "^2.0.0",
     "cross-env": "^7.0.0",
+    "debug": "^4.0.0",
     "dtslint": "^4.0.0",
     "eslint-plugin-es": "^4.0.0",
     "eslint-plugin-security": "^1.0.0",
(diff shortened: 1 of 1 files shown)


TrySound (approved):

wooorm:
Thanks for the PR, but I don’t think it should be done. Although, the fact that this arose sounds like micromark is confusing, so it might be good to discuss if there are alternatives.

There are two copies published: `lib/` and `dist/`. `lib/` is the source code, which is instrumented with assertions and debugging. `dist/` is basically the same, but the assertions and debugging is removed (and constants are inlined). Both run fine everywhere, but assertions and debugging is mostly useful when writing extensions. A different use case is if there is a bug, then you can switch to use `micromark/lib` instead of `micromark`. But generally, it’s fine to use the smaller bundle.

This difference is [documented in the readme](https://github.com/micromark/micromark#size--debug): `debug` and `assert` are used in some of the code that (might) run, so in my opinion are dependencies.

One alternative to this approach is what facebook/react does: [a `__DEV__` variable](https://github.com/facebook/react/blob/e6a0f276307fcb2f1c5bc41d630c5e4c9e95a037/packages/react/src/ReactElement.js#L27), but that’s a lot of complexity and requires a bundler.

I’d love to hear other ideas on how to make this better.

TrySound:
AFAIK node and bundlers considering to support development/production fields for "exports" like this. Though need to check
```
{
  exports: {
    ".": {
      development: "...",
      production: "...",
      import: "...",
      require: "..."
    }
  }
}
```

fazouane-marouane:
hum I always wondered what would be shipped in npm. Thanks for the clarifications. If we also ship "./lib" then I have no objection in keeping debug package in the dependencies ! Shall I close this PR  ?

wooorm:
done!

@TrySound Would love to read more about that

PR #42, opened 2020-12-09
Add typing for Construct.name

ocavue:
This PR adds typing for commit 71561f3.

### DIFF
--- a/lib/shared-types.d.ts
+++ b/lib/shared-types.d.ts
@@ -165,6 +165,7 @@ export type Resolve = (events: Event[], context: Tokenizer) => Event[]
 export type Tokenize = (context: Tokenizer, effects: Effects) => State
 
 export interface Construct {
+  name?: string
   tokenize: Tokenize
   partial?: boolean
   resolve?: Resolve


lib/shared-types.d.ts
```diff
@@ -165,6 +165,7 @@ export type Resolve = (events: Event[], context: Tokenizer) => Event[]
 export type Tokenize = (context: Tokenizer, effects: Effects) => State
 
 export interface Construct {
+  name?: string
```
ChristianMurphy, line 168: this is a good starting point,
down the road it may make sense to enumerate the different construct names and/or make this generic to fix the construct name

ChristianMurphy (approved):
Thanks @ocavue !

lib/shared-types.d.ts
wooorm: some more background here: `name` as I implemented it can be anything. When it’s set, constructs can be turned off. Maybe extensions would also like that, but then again: turning constructs off only really makes sense for core constructs.

But maybe there are use cases for future things (debugging?) where `name` would be useful for all constructs

wooorm (approved):

wooorm:
Released!

PR #44, opened 2020-12-13
Refactor to fix misc small things

fazouane-marouane:
I tried sonarcloud and deepscan to see if there are some small fixes to be done.

This PR fixes:
- a condition `code > 55295` checked twice
- exported types may contain duplicate type literals. Also I wonder if exporting the type "Constant" is good since it contains heterogeneous data. Maybe it just needs splitting ?
- `@rollup/plugin-babel` seems to support ESM as of recently

Should we also include sonarcloud/deepscan (or maybe other tools) to this repo so the main branch and futur PRs could be checked as well ?


lib/util/normalize-uri.mjs
```diff
@@ -38,7 +38,7 @@ function normalizeUri(value) {
       next = value.charCodeAt(index + 1)
 
       // A correct surrogate pair.
-      if (code > 55295 && code < 56320 && next > 56319 && next < 57344) {
+      if (code < 56320 && next > 56319 && next < 57344) {
```
wooorm, line 41: nice!

wooorm (approved):
Nice!

wooorm:
> Maybe it just needs splitting ?

Can you clarify what you mean here?

> Should we also include sonarcloud/deepscan (or maybe other tools) to this repo so the main branch and futur PRs could be checked as well ?

How would that work? Would they be added as a GH action?

ChristianMurphy (approved):

fazouane-marouane:
> > Maybe it just needs splitting ?
> 
> Can you clarify what you mean here?

I was referring to constants.d.ts exporting Constants as one single type while we can maybe not generate the Constants.d.ts file since it's not meant to be used by code that depends on micromark, if my understanding is correct.

> > Should we also include sonarcloud/deepscan (or maybe other tools) to this repo so the main branch and futur PRs could be checked as well ?
> 
> How would that work? Would they be added as a GH action?

I tried them only by signing-in their website. They are apps that you can install on a repo and they will auto-subscribe to github hooks (like dependabot for example). I also tried https://lgtm.com which tests for security issues among other things and everything looks good on micromark 👍

rollup.config.mjs
```diff
@@ -1,14 +1,9 @@
...
 import commonjs from '@rollup/plugin-commonjs'
 import {nodeResolve} from '@rollup/plugin-node-resolve'
 import {terser} from 'rollup-plugin-terser'
 
-// eslint-disable-next-line node/no-deprecated-api -- Remove when `@rollup/plugin-babel` supports ESM.
-var requireUtil = module.createRequireFromPath(
-  path.join(process.cwd(), 'rollup.config.mjs')
-)
-var babel = requireUtil('@rollup/plugin-babel').babel
+import {babel} from '@rollup/plugin-babel'
```
TrySound, line 6: Did you try to run test? I have no idea why CI passed this but babel plugin does not have es modules support. `module` field is only for bundlers. Node does not use it.

TrySound: Please revert.

TrySound: Oh, I see. It's not babel that supports es modules. But node now can have named imports from commonjs. And it's backported to latest node 12 and 14. Very interesting.

wooorm: Wait that sounds interesting @TrySound, what did you find out?

TrySound: I had a few month old node which failed with this change. After update it worked.
Looks like node team implemented named exports detection in commonjs https://github.com/nodejs/node/pull/35249

PR #105, opened 2021-12-12
Fix to generate a bundle for fuzzing

ChristianMurphy:
### Initial checklist

### Description of changes

Pending a fix for https://gitlab.com/gitlab-org/security-products/analyzers/fuzzers/jsfuzz/-/issues/5
coverage guidance is currently limited to cjs in jsfuzz, this generates a temporary bundle file using esbuild to allow for coverage guidance to speed up the fuzzing process.


package.json
```diff
@@ -90,7 +91,7 @@
     "build-size": "rollup -c --silent && gzip-size micromark.min.js && gzip-size --raw micromark.min.js",
     "build": "npm run build-workspace && npm run build-monorepo && npm run build-size",
     "format": "remark . -qfo && prettier . -w --loglevel warn && xo --fix",
-    "pretest-fuzz": "patch-package --patch-dir script/patches && node script/generate-fixtures.js",
+    "pretest-fuzz": "patch-package --patch-dir script/patches && echo \"import {micromark} from 'micromark'; export {micromark}\" | esbuild --format=cjs --platform=node --bundle > test/fuzz-bundle.cjs && node script/generate-fixtures.js",
```
wooorm, line 94: `import {micromark} from 'micromark'; export {micromark}` seems unneeded? can’t the source be bundled? 🤔

ChristianMurphy: I use this as a short hand to find the right file which should be bundled.
Following node's folder rules.

It could be pointed at a specific file, but then we'd need to hard code a specific (potentially changing) path.

wooorm: Doesn't esbuild look into node_modules, and nom workspaces put mm there tho?

ChristianMurphy: right, but it needs an entrypoint, and I have not found an option to bundle just a specific module yet, so I create a sample file to start that process.

wooorm: Oh I though esbuild did allow "foo" if it was in node_modules! OK!

ChristianMurphy: Maybe?
Trying to do a clean build and test this now and discovered npm is down. https://status.npmjs.org/incidents/4p6mgfgs0nxt
May have to wait to test

wooorm: Expand the details here: https://mdxjs.com/docs/troubleshooting-mdx/#esm.
That’s picking up `node_modules/@mdx-js/mdx`, hence when I was guessing that an extra file wouldn’t be needed?

ChristianMurphy: updated in 1336066

packages/micromark-util-character/dev/lib/unicode-punctuation-regex.js
```diff
@@ -4,5 +4,4 @@
...
...BE\u05C0\u05C3\u05C6\u05F3\u05F4\u0609\u060A\u060C\u060D\u061B\u061D-\u061F\u066A-\u066D\u06D4\u0700-\u070D\u07F7-\u07F9\u0830-\u083E\u085E\u0964\u0965\u0970\u09FD\u0A76\u0AF0\u0C77\u0C84\u0DF4\u0E4F\u0E5A\u0E5B\u0F04-\u0F12\u0F14\u0F3A-\u0F3D\u0F85\u0FD0-\u0FD4\u0FD9\u0FDA\u104A-\u104F\u10FB\u1360-\u1368\u1400\u166E\u169B\u169C\u16EB-\u16ED\u1735\u1736\u17D4-\u17D6\u17D8-\u17DA\u1800-\u180A\u1944\u1945\u1A1E\u1A1F\u1AA0-\u1AA6\u1AA8-\u1AAD\u1B5A-\u1B60\u1B7D\u1B7E\u1BFC-\u1BFF\u1C3B-\u1C3F\u1C7E\u1C7F\u1CC0-\u1CC7\u1CD3\u2010-\u2027\u2030-\u2043\u2045-\u2051\u2053-\u205E\u207D\u207E\u208D\u208E\u2308-\u230B\u2329\u232A\u2768-\u2775\u27C5\u27C6\u27E6-\u27EF\u2983-\u2998\u29D8-\u29DB\u29FC\u29FD\u2CF9-\u2CFC\u2CFE\u2CFF\u2D70\u2E00-\u2E2E\u2E30-\u2E4F\u2E52-\u2E5D\u3001-\u3003\u3008-\u3011\u3014-\u301F\u3030\u303D\u30A0\u30FB\uA4FE\uA4FF\uA60D-\uA60F\uA673\uA67E\uA6F2-\uA6F7\uA874-\uA877\uA8CE\uA8CF\uA8F8-\uA8FA\uA8FC\uA92E\uA92F\uA95F\uA9C1-\uA9CD\uA9DE\uA9DF\uAA5C-\uAA5F\uAADE\uAADF\uAAF0\uAAF1\uABEB\uFD3E\uFD3F\uFE10-\uFE19\uFE30-\uFE52\uFE54-\uFE61\uFE63\uFE68\uFE6A\uFE6B\uFF01-\uFF03\uFF05-\uFF0A\uFF0C-\uFF0F\uFF1A\uFF1B\uFF1F\uFF20\uFF3B-\uFF3D\uFF3F\uFF5B\uFF5D\uFF5F-\uFF65]/
```
ChristianMurphy, line 7: this was automatically updated when running `npm run build`

package.json
wooorm: ah nice, so it does work, right? Perfect! :+1:

packages/micromark-util-character/dev/lib/unicode-punctuation-regex.js
wooorm: an `npm format` should change it again! :+1:

ChristianMurphy: updated

wooorm (approved):

wooorm:
Sweeeet!

PR #106, opened 2021-12-12
Canary test workflow, checks that key downstream packages still build

ChristianMurphy:
### Initial checklist

### Description of changes

Follow up to: https://github.com/unifiedjs/ideas/issues/7

> Unified is a large and highly connected community of packages.
> Changes to any package could have impact downstream packages.
> Currently anticipating impact is a purely manual process.

> It could be helpful to use downstream packages, which themselves have comprehensive test suites, to help automatically detect what changes may be breaking or have unintended impacts.

This aims to help alleviate some of that risk by choosing a few key packages.
* mdast-util-from-markdown
* remark
* react-markdown

which build on this, and have heavy usage, and comprehensive test suites, to anticipate potential downstream issues faster.


.github/workflows/canary.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: checkout mdast-util-from-markdown
+        uses: actions/checkout@v2
+        with:
+          repository: syntax-tree/mdast-util-from-markdown
+          path: canary/mdast-util-from-markdown
+      - name: test mdast-util-from-markdown
+        run: |
+          npm install
+          npx rimraf "**/micromark"
+          npx rimraf "**/micromark-util-*"
```
ChristianMurphy, line 27: this removes the pinned versions of these libraries that npm installed, meaning node will continue searching upward, until it finds the parent `micromark/node_modules` folder with the just-built version of micromark.

ChristianMurphy:
Good news: this works!
Bad news: there appear to be failing tests upstream

.github/workflows/canary.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: checkout ${{matrix.package}}
+        uses: actions/checkout@v2
+        with:
+          repository: ${{matrix.package}}
+          path: canary/${{matrix.package}}
+      - name: test ${{matrix.package}}
+        run: |
+          npm install
+          npx rimraf "**/micromark"
```
wooorm, line 31: Wouldn’t `npx` significantly slow down this process compared to having it (guessing we already have it?)

wooorm (commented):
Fixing the CI failure is most important, otherwise, good idea!

wooorm:
Perhaps something to not run this on every PR? Similar to that “T-Shaped” model from MDX?

ChristianMurphy:
> Perhaps something to not run this on every PR?

I think there would be value in running this on every PR, seeing if code changes impact upstream.
And even if changes aren't code related, it can be an early warning for unrelated upstream issues.
This set of checks itself probably wouldn't be blocking for PRs.

> Similar to that “T-Shaped” model from MDX?

There the T-Shaped model has to do with the matrix of jobs run, not when it is run, right?
I did drop windows from the matrix for now, because it appears most upstream projects do not test on Windows.
Which (may) eliminate the need to a T-Shaped model.

.github/workflows/canary.yml
ChristianMurphy: We already have it, `npx` is allowing us to call it without needing to specify which local `node_modules` folder it is in.

.github/workflows/canary.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    strategy:
+      fail-fast: false
+      max-parallel: 2
+      matrix:
+        os:
+          - ubuntu-latest
+        node:
+          - lts/erbium
+          - node
```
wooorm, line 44: I see no strong benefit in trying to test Erbium, assuming that in almost all cases the result would be the same as on `node`?

.github/workflows/canary.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: setup node
+        uses: dcodeIO/setup-node-nvm@master
+        with:
+          node-version: ${{matrix.node}}
+      - name: update npm
+        run: npm install -g npm
+      - name: install micromark
+        run: npm install
+      - name: build micromark
+        run: npm run build
```
wooorm, line 21:
```suggestion
      - name: micromark
        run: npm install -g npm && npm install && npm run build
```

This should likely work and be uninteresting. And if it doesn’t work, the other CIs should fail

.github/workflows/canary.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: checkout ${{matrix.package}}
+        uses: actions/checkout@v2
+        with:
+          repository: ${{matrix.package}}
+          path: canary/${{matrix.package}}
+      - name: test ${{matrix.package}}
+        run: |
+          npm install
+          npx rimraf "**/micromark"
+          npx rimraf "**/micromark-util-*"
```
wooorm, line 32: What is being removed here? Are there `micromark` files or folders in `micromark/canary/remarkjs/remark/`?
Why just `micromark` itself and `micromark-util-*`, but not other `micromark-*` files/folders?

.github/workflows/canary.yml
```diff
@@ -0,0 +1,48 @@
+name: canary
+on:
+  - pull_request
+  - push
+jobs:
+  canary:
+    name: testing ${{matrix.package}} with ${{matrix.node}} on ${{matrix.os}}
```
wooorm, line 7: (see later for more info, but: I think testing only on `ubuntu-latest` and `node` is already useful, and I don’t see much added benefits by checking older node too).

```suggestion
    name: ${{matrix.package}}
```

Name is quite long and can’t be displayed properly (sidebar here: https://github.com/micromark/micromark/runs/4620836214?check_suite_focus=true)

.github/workflows/canary.yml
wooorm: Can also use `|`s for multilines, like the next item

.github/workflows/canary.yml
ChristianMurphy: copying from earlier in the discussion https://github.com/micromark/micromark/pull/106#discussion_r767339359

> this removes the pinned versions of these libraries that npm installed, meaning node will continue searching upward, until it finds the parent `micromark/node_modules` folder with the just-built version of micromark.

---

> Are there micromark files or folders in micromark/canary/remarkjs/remark/?

There are, this runs right after `npm install` in the _micromark/canary/remarkjs/remark_ folder, so it has a frozen/published copy of micromark installed, we want it to find the latest, just built, copy instead.

> Why just micromark itself and micromark-util-*, but not other micromark-* files/folders?

It seemed to work, open to suggestions for a better remove pattern to get all the micromark packages in this repo cleared at once.

wooorm: ahhh, forgot that I already asked it.
Wouldn’t `npx rimraf "node_modules/micromark*"` do the same but catch everything in [`packages/`](https://github.com/micromark/micromark/tree/main/packages)? (or `npx rimraf "node_modules/{micromark,micromark*}"` if the asterisk matches one or more, which I don’t think it does?)

ChristianMurphy: One possible alternative, something like

```
rm -rf `npm ls --workspaces --json | jq "[.dependencies | keys[]] | \"**/\" + .[]"`
```

which tries to get at the parent workspace, list the packages installed there, create a `**/{package name}` pattern for each, then passing all that back to `rm -rf`

wooorm: Hmm, but what about packages needed in the canary packages though? `micromark-extension-mdx`?
Problem is that this repo houses `micromark-factory-*` too, but those extensions can also use `micromark-*` packages? 🤔

ChristianMurphy: > Wouldn’t `npx rimraf "node_modules/micromark*"` do the same but catch everything in packages/?

it would! but, it may be too aggressive, also removing things like `micromark-extension-gfm` which may be needed by unit tests and not part of the parent micromark repo.

wooorm: Would `rm -rf "canary/.../node_modules/${ls packages/}"` do the same, but cleaner?

ChristianMurphy: which admittedly can even happen with my current pattern.
For example

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'micromark-util-events-to-acorn' imported from /home/runner/work/micromark/micromark/canary/micromark/micromark-extension-mdx/node_modules/micromark-factory-mdx-expression/dev/index.js
    at new NodeError (node:internal/errors:371:5)
    at packageResolve (node:internal/modules/esm/resolve:864:9)
    at moduleResolve (node:internal/modules/esm/resolve:910:18)
    at defaultResolve (node:internal/modules/esm/resolve:1005:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:530:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:251:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:79:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

https://github.com/micromark/micromark/runs/4627735817?check_suite_focus=true#step:6:36

wooorm:
Looking at the, how do you call this, build preview, it looks chaotic imo.
- Could the canary/* stuff be ordered after main/*?
- `canary / testing / $user/$project` -> `canary-in-the-gold-mine / $package` or `citgm / $package` maybe?

Screen Shot 2021-12-24 at 16 46 09

.github/workflows/canary.yml
ChristianMurphy: > Would `rm -rf "canary/.../node_modules/${ls packages/}"` do the same, but cleaner?

potentially, I think we want the reverse of this though. :thinking: 
inside the canary folder, we want to list the packages in the parent micromark folder, and use that list to remove frozen/published copies in the canary folder.

ChristianMurphy:
> Could the canary/* stuff be ordered after main/*?

I'm not sure if it can.
The order appears to be non-alphabetical.

> `canary / testing / $user/$project` -> `canary-in-the-gold-mine / $package or citgm` / `$package` maybe?

Happy to rename things if wanted, I'm not sure it will fix the order issue though.

wooorm:
> I'm not sure if it can. The order appears to be non-alphabetical.

Yeah, though, taking MDX as an example: https://github.com/mdx-js/mdx/pull/1854, it seems to indicate that different YAML files result in unordered, but when inside the same file, it does work?

> Happy to rename things if wanted, I'm not sure it will fix the order issue though.

I meant the points separately. I guess `$packageName` doesn’t make sense though, but `$user/$project` does, as this seems to be more about the repo rather the npm package?

.github/workflows/canary.yml
wooorm: Huh, that’s what my suggestion does, right?
- List all packages maintained in the micromark monorepo (which are hoisted to the monorepo `node_modules/`?)
- Install an example repo normally, separately from the monorepo
- Remove what monorepo packages from the example repo `node_modules`

wooorm: Though my example was assuming a CWD of the monorepo, from an example repo my example would be, say, `rm -rf "./node_modules/${ls ../../../packages/}"`

ChristianMurphy: testing locally I get `bad substitution`.
But parsing out your intent.

This script is running inside the canary/{package-name} folder.
so `ls packages` would list package inside the `canary` folder right?
`canary` would then try to go into a folder that doesn't exist.
`...` would take it up two levels to the parent micromark folder
`node_modules` would go into the parent's `node_modules` directory

ChristianMurphy:
> but when inside the same file, it does work?

Moved canary job into main.yml

> as this seems to be more about the repo rather the npm package?

it is more about the repositories, updated naming to better reflect this

.github/workflows/canary.yml
ChristianMurphy: resolved in  42e5976

wooorm (approved):

PR #108, opened 2022-01-10
Add a community plugin

leebyron:
### Initial checklist

### Description of changes

https://github.com/leebyron/remark-comment

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -287,6 +287,9 @@ As a (potential) author of extensions, refer to
 
 *   [`wataru-chocola/micromark-extension-definition-list`](https://github.com/wataru-chocola/micromark-extension-definition-list)
     — support definition lists
+*   [`leebyron/remark-comment`](https://github.com/leebyron/remark-comment)
+    — includes micromark plugin for omitting HTML comments from rendered HTML
+
 
 ### `SyntaxExtension`
 


wooorm:
Hey nice!

Hmm, I’m not too sure about using the label “remark” here, it looks off, I’m guessing that it would confuse folks 🤔
I do understand that 4 repos is too much, though maybe `packages/` might work. Or maybe I should just make up my mind and not see it as a problem 🤷‍♂️

for the description, maybe:

```diff
-includes micromark plugin for omitting HTML comments from rendered HTML
+includes micromark extension for HTML comments, useful when using MDX syntax
```

Or:

```diff
-includes micromark plugin for omitting HTML comments from rendered HTML
+includes micromark extension for HTML comments, helpful when migrating to MDX 2
```

…or so?

---

Two more points:
- There are some other patterns in comments not allowed (according to HTML, which is different from HTML), which you might find interesting or useful: https://github.com/syntax-tree/hast-util-to-html/blob/919d63bb990983d12a4270ce5d16e7b803512c32/lib/comment.js#L21
- HTML comments were removed in v2 btw, I’m assuming you know this, but wanted to make sure you did

wooorm:
Ping! :)

PR #122, opened 2022-10-04
feat: add definitions to CompileDataFields

### LINKED ISSUE #121: Make `definitions` available to extensions
### Initial checklist

- [X] I read the [support docs](https://github.com/micromark/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/micromark/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/micromark/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amicromark&type=Issues) and couldn’t find anything (or linked relevant results below)

### Problem

i'm writing an extension where I would need the [definitions](https://github.com/micromark/micromark/blob/27d3426f0181215a60700424c36671467592e495/packages/micromark/dev/lib/compile.js#L84) defined in the document. 

### Solution

the definitions should be available either via the context `this.definitions` or via `getData('mediaDefinitions')`.
the latter is probably better.

### Alternatives

I could probably overwrite all the definition related enter and exit methods, and track them as well, but this sounds like a wrong approach.

tripodsan:
fixes #121

### Initial checklist

### Description of changes

Added the `definitions` records that track the media definitions in the markdown to the compiler data object. This allows extensions to make use of them.

### DIFF
--- a/packages/micromark-util-types/index.js
+++ b/packages/micromark-util-types/index.js
@@ -496,6 +496,10 @@
  * @typedef {Partial<NormalizedHtmlExtension>} HtmlExtension
  *   An HTML extension changes how markdown tokens are serialized.
  *
+ * @typedef Definition
+ * @property {string} [destination]
+ * @property {string} [title]
+ *
  * @typedef _CompileDataFields
  * @property {boolean} lastWasTag
  * @property {boolean} expectFirstItem
@@ -509,6 +513,7 @@
  * @property {boolean} inCodeText
  * @property {string} characterReferenceType
  * @property {Array<boolean>} tightStack
+ * @property {Record<string, Definition>} definitions
  *
  * @typedef {Record<string, unknown> & Partial<_CompileDataFields>} CompileData
  *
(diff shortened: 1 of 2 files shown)


tripodsan:
@wooorm do you know why the react-markdown test fails? it works locally.

wooorm:
@tripodsan Did you test this out locally (in `node_modules`)? Does it work for you?

tripodsan:
> @tripodsan Did you test this out locally (in `node_modules`)? Does it work for you?

what do you mean? I used the PR version in my extension and it would solve my problem - if this is what you mean.

wooorm:
It would, or it did?
I'm trying to make sure this works for you. So I don't need to do another release soon to actually solve it some other way. Or to later find out that it's not enough information for you to solve your problem.

tripodsan:
yes. the proposed solution works for my use case. thank you.

wooorm:
Released in [3.1.0](https://github.com/micromark/micromark/releases/tag/3.1.0)!

PR #124, opened 2022-10-17
Fix some tests not actually running

shartte:
### Initial checklist

### Description of changes

I noticed that the commonmark test only runs the last section from the commonmark test cases ('Textual content') and ignores the rest. This is due to `key` being modified after being captured.

After this change, running commonmark.js individually yields:
```
1..652
# tests 652
# pass  652
```

Without the patch, it printed:
```
1..78
# tests 78
# pass  78
```

### DIFF
--- a/test/commonmark.js
+++ b/test/commonmark.js
@@ -23,14 +23,12 @@ test('commonmark', function (t) {
   for (key in sections) {
     if (!hasOwnProperty.call(sections, key)) continue
 
+    const section = sections[key]
     t.test(key, function (t) {
       let index = -1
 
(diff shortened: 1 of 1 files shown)


wooorm:
Oh whoops. Must be due to some refactoring, because I’m pretty sure it used to work and it’s all still CommonMark-compliant.

@shartte Can you add the initial checklist to your PR body? Thanks

ChristianMurphy (approved):

wooorm (approved):

wooorm:
Thanks @shartte!

PR #132, opened 2023-01-17
Fix internal types for TS4.9, add `module: node16`

ChristianMurphy:
### Initial checklist

*   [x] I read the support docs 
*   [x] I read the contributing guide 
*   [x] I agree to follow the code of conduct 
*   [x] I searched issues and couldn’t find anything (or linked relevant results below) 
*   [x] If applicable, I’ve added docs and tests

### Description of changes

This includes:
- [x] TypeScript 4.9 fixes
- [x] Internal fixes TypeScript ESM resolution in Node16 mode
- [x] New conditional exports for TypeScript ESM

PR #133, opened 2023-01-17
Update `rollup` to v3

ChristianMurphy:
### Initial checklist

### Description of changes

rollup 3 uses @types/estree 1+ which resolves a number of library eslint warnings currently triggered by pinning 0.x

Changelog: https://github.com/rollup/rollup/blob/master/CHANGELOG.md#300

We already use Node 14+.
Rollup using `node:` protocol imports sounds like an improvement.
Configuration changes look like options we're not using.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
   ],
   "devDependencies": {
     "@rollup/plugin-node-resolve": "^15.0.0",
-    "@rollup/plugin-terser": "^0.3.0",
+    "@rollup/plugin-terser": "^0.4.0",
     "@types/concat-stream": "^2.0.0",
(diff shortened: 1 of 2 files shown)


ChristianMurphy:
depends on #132 for TS3.9 fixes

wooorm:
This seems to be broken even with recent changed, not sure what’s up?

ChristianMurphy:
It looks like the import moved from a named to a default.
Fixed in [ce54772](https://github.com/micromark/micromark/pull/133/commits/ce547723ddef890334763bf6eec85fe4c6c4544e)

wooorm (approved):

wooorm:
thanks!

Issue #131, opened 2023-01-14
TokenizeContext.sliceSerialize throws in sliceChunks if first chunk of token is Code instead of string


DavidAnson:
### Initial checklist

- [X] I read the [support docs](https://github.com/micromark/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/micromark/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/micromark/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amicromark&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

micromark@3.1.0

### Steps to reproduce

```bash
user@HOST micromark-issue % npm ls micromark
micromark-issue@ /Users/user/micromark-issue
└── micromark@3.1.0

user@HOST micromark-issue % cat issue.mjs 
import { parse } from "micromark/lib/parse";
import { postprocess } from "micromark/lib/postprocess";
import { preprocess } from "micromark/lib/preprocess";

function repro(markdown) {
  console.log("trying...");
  const encoding = undefined;
  const end = true;
  const options = undefined;
  const chunks = preprocess()(markdown, encoding, end);
  const parseContext = parse(options).document().write(chunks);
  const events = postprocess(parseContext);
  for (const event of events) {
    const [ \_, token, context ] = event;
    context.sliceSerialize(token);
  }
  console.log("ok");
}

repro("Heading\\n=======");
repro("\\nHeading\\n=======");
user@HOST micromark-issue % node issue.mjs 
trying...
ok
trying...
file:///Users/user/micromark-issue/node\_modules/micromark/lib/create-tokenizer.js:520
      view[0] = view[0].slice(startBufferIndex)
                        ^

TypeError: view[0].slice is not a function
    at sliceChunks (file:///Users/user/micromark-issue/node\_modules/micromark/lib/create-tokenizer.js:520:25)
    at sliceStream (file:///Users/user/micromark-issue/node\_modules/micromark/lib/create-tokenizer.js:154:12)
    at Object.sliceSerialize (file:///Users/user/micromark-issue/node\_modules/micromark/lib/create-tokenizer.js:149:28)
    at repro (file:///Users/user/micromark-issue/issue.mjs:15:13)
    at file:///Users/user/micromark-issue/issue.mjs:21:1
    at ModuleJob.run (node:internal/modules/esm/module\_job:198:25)
[...]

PR #137, opened 2023-03-02
Fix exception in internal `sliceSerialize` API

DavidAnson:
Closes GH-131.

### Initial checklist

### Description of changes

Fixes a problem where TokenizeContext.sliceSerialize throws in sliceChunks if the first chunk of token is Code instead of string. See #131 for discussion. All code here was written by @wooorm.


wooorm (approved):

wooorm:
Thanks! I’ll wait a bit before releasing this, let’s see if we can get some more changes in first!

PR #138, opened 2023-03-03
chore: update github actions to latest version

ChristianMurphy:
### Initial checklist

### Description of changes

upgrades check action to latest major.
upgrade codecov action to latest major.
migrate from nvm action to setup-node action (which now supports named releases).


ChristianMurphy:
@wooorm @remcohaszing thoughts on [b29a5e6](https://github.com/micromark/micromark/pull/138/commits/b29a5e6c90a0044990c1ad68ae310441208352c6)?
I ran into https://github.com/npm/cli/issues/4341, I added the jump to 8.3.1 as a work around, in order to keep the test matrix the same.
But we could also drop `lts/fermium` (node 14) from the matrix, and add `lts/hydrogen` (node 18).

remcohaszing:
The duplicate `npm install` isn’t pretty, but if it’s needed, then so be it.

You could make this step conditional.

```yaml
- if: ${{matrix.node-version == 'lts/fermium' && matrix.os == 'windows-latest'}}
  run: |
    npm install -g npm@8.3.1
    npm install -g npm@latest
```

Perhaps upgrading to latest isn’t even needed.

wooorm:
ehh, that doesn’t look great. Wasn‘t that npm problem fixed? I’ve been able to use Windows in all `vfile/*` repos (which do use `lts/gallium`, [example](https://github.com/vfile/vfile-message/blob/main/.github/workflows/main.yml))

ChristianMurphy:
> Wasn‘t that npm problem fixed? 

It was/is fixed, but `lts/fermium` didn't get the fix.

> Maybe switch to lts/gallium and then node instead?

Sounds good, incoming

ChristianMurphy:
[dccf55d](https://github.com/micromark/micromark/pull/138/commits/dccf55d9d4a38335de135374e15089a8eee8652b) drops LTS Fermium and adds LTS Gallium as well as latest `node`, removing the `npm` quirk.

remcohaszing (approved):

wooorm (approved):

wooorm:
Thank you!!

PR #142, opened 2023-03-16
Update `typescript`

ChristianMurphy:
### Initial checklist

### Description of changes

TypeScript 4 missed that some interfaces weren't explicitly imported. TypeScript 5 requires them.
TypeScript 5 is unable to infer the inline html extensions. Explicitly noting they are extensions seems to make the issue go away. Other suggestions are welcome.


wooorm:
Thanks!

PR #143, opened 2023-03-17
Add `exactOptionalPropertyTypes` to `tsconfig.json`

ChristianMurphy:
### Initial checklist

### Description of changes

This enables tighter definition on what can or cannot be explicitly assigned `undefined`.
Currently I have the minimum number of `| undefined` needed to make tests pass.

`packages/micromark-util-types/index.js` is the file I'd draw the most attention to.
There are other uses of optional properties where I did not add `| undefined` yet, I'm unsure if there is intent for them to allow explicit `undefined`.

### DIFF
--- a/packages/micromark-util-resolve-all/index.js
+++ b/packages/micromark-util-resolve-all/index.js
@@ -7,7 +7,7 @@
 /**
  * Call all `resolveAll`s.
  *
- * @param {Array<{resolveAll?: Resolver}>} constructs
+ * @param {Array<{resolveAll?: Resolver | undefined}>} constructs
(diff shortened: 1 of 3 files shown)


wooorm (approved):
Good to go, right?

ChristianMurphy:
> Good to go, right?

I think so.
My only question/concern was

> `packages/micromark-util-types/index.js` is the file I'd draw the most attention to.
There are other uses of optional properties where I did not add `| undefined` yet, I'm unsure if there is intent for them to allow explicit `undefined`.

So long as we know those are _not_ intended to allow explicit `undefined` we're good.

PR #158, opened 2023-10-26
Fix handling of astral character reference

Porges:
### Initial checklist

### Description of changes

`decodeNumericCharacterReference` was using `String.fromCharCode` which only handles UTF-16 code units, not true codepoints. Any characters not representable by a single UTF-16 code unit ("astral characters") would come out incorrectly.

Note that there are other uses of `String.fromCharCode` in micromark. These should also be examined. The current behaviour is a regression that was introduced 4 months ago by bbb726d240ab8736daf5276555d17a6292e43dae, although the correct version only existed for 3 days prior to that 🙂 — the very first version of the function had the same problem.

### DIFF
--- a/packages/micromark-util-decode-numeric-character-reference/dev/index.js
+++ b/packages/micromark-util-decode-numeric-character-reference/dev/index.js
@@ -38,5 +38,5 @@ export function decodeNumericCharacterReference(value, base) {
     return values.replacementCharacter
   }
 
-  return String.fromCharCode(code)
+  return String.fromCodePoint(code)
 }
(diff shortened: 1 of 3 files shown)


wooorm (approved):

wooorm:
Thanks!

wooorm:
Our linter `xo` thought it was useful. But we mostly index into strings, so changing to code points actually breaks things. In this case it doesn’t though!

wooorm:
Thank you, released! https://github.com/micromark/micromark/releases/tag/micromark-util-decode-numeric-character-reference%402.0.1

Porges:
> Thank you, released! https://github.com/micromark/micromark/releases/tag/micromark-util-decode-numeric-character-reference%402.0.1

Thanks for the quick turnaround

PR #165, opened 2024-03-24
Add license file to micromark package directory so it's included in NPM

paulbarmstrong:
### Initial checklist

### Description of changes
This copies the license file from the root repo directory into the packages/micromark directory so that it's included in NPM. I believe projects' licenses are typically included in NPM. That makes it easier to satisfy this license's requirement for it to be included in all copies or substantial portions of the Software.

Thanks in advance for reviewing the PR. I didn't see any GitHub issues that discuss the situation with the license in NPM. I hope this change seems worthwhile.

### DIFF
--- a/packages/micromark/license
+++ b/packages/micromark/license
@@ -1,0 +1,22 @@
+(The MIT License)
+
+Copyright (c) 2020 Titus Wormer <tituswormer@gmail.com>
+
+Permission is hereby granted, free of charge, to any person obtaining
+a copy of this software and associated documentation files (the
+'Software'), to deal in the Software without restriction, including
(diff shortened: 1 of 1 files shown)


paulbarmstrong:
I think there might currently be a problem with the tip of `micromark:main`. I encountered the same `Error: test/io/misc/zero.js(1,1): error TS1490: File appears to be binary.` error from the build above when I tried to build it locally.

paulbarmstrong:
I realized that this issue affects all the NPM packages of the packages of this monorepo. I'll close this since I think this may be best solved by a more complicated mechanism.

PR #166, opened 2024-03-24
Replace use of literal replacement character in test code

paulbarmstrong:
### Initial checklist

### Description of changes

[Currently the tip of `micromark:main` fails to build due to: `test/io/misc/zero.js:1:1 - error TS1490: File appears to be binary.`](https://github.com/micromark/micromark/actions/runs/8411681502/job/23031546103?pr=165) I'm not able to find documentation about how the TS complier decides to throw this error but it seems to depend on a combination of factors including the presence of the replacement character. This change addresses that error by replacing all uses of the replacement character in the test code with the equivalent escape sequence.

Replacing just the first use of the replacement character in zero.js allows `npm test` to work but I replace it in all test code to prevent this issue elsewhere if the TS complier's criteria changes (as it seems to have done since the last time `micromark:main` was able to build). I leave it as-is in the non-test source code since I think that code is built and used in a more sophisticated way and I don't want to unintentionally alter the behavior.


wooorm:
Heya! I wonder if this can be considered a bug in TS, or is working as intended. Checking here: https://github.com/microsoft/TypeScript/pull/57008#issuecomment-2017698194. I will report back here when I hear from them there.

wooorm:
Thanks! Went with a tiny change (a comment describing why the comment exists) as I believe this will be fixed in TS in the future.

Issue #169, opened 2024-03-26
Performance: larger MDX files are unmanagably slow to parse


robsimmons:
### Initial checklist

- [X] I read the [support docs](https://github.com/micromark/.github/blob/main/support.md)
- [X] I read the [contributing guide](https://github.com/micromark/.github/blob/main/contributing.md)
- [X] I agree to follow the [code of conduct](https://github.com/micromark/.github/blob/main/code-of-conduct.md)
- [X] I searched [issues](https://github.com/search?q=user%3Amicromark&type=Issues) and couldn’t find anything (or linked relevant results below)

### Affected packages and versions

mdast-util-from-markdown 2.0.0, micromark-extension-mdjs 3.0.0

### Link to runnable example

[...]

PR #170, opened 2024-03-26
Refactor to improve performance w/ push, reverse

robsimmons:
### Initial checklist

### Description of changes

This incompletely addresses the slow loop causing #169. In my testing, about 10% of the time spent in that loop is repeatedly shifting data onto the front of a large array that is definitely empty at the start of the loop. The V8 engine, at least, has bad performance for shift/unshift relative to push/pop.

### DIFF
--- a/packages/micromark-util-subtokenize/dev/index.js
+++ b/packages/micromark-util-subtokenize/dev/index.js
@@ -246,10 +246,11 @@ function subcontent(events, eventIndex) {
     const slice = childEvents.slice(breaks[index], breaks[index + 1])
     const start = startPositions.pop()
     assert(start !== undefined, 'expected a start position when splicing')
-    jumps.unshift([start, start + slice.length - 1])
+    jumps.push([start, start + slice.length - 1])
     splice(events, start, 2, slice)
   }
 
+  jumps.reverse()
   index = -1
 
   while (++index < jumps.length) {


wooorm:
Yeah, in the rust version I too have started reversing a lot.

wooorm (approved):

wooorm:
Thanks!

packages/micromark-util-subtokenize/dev/index.js
```diff
@@ -246,10 +246,11 @@ function subcontent(events, eventIndex) {
     const slice = childEvents.slice(breaks[index], breaks[index + 1])
     const start = startPositions.pop()
     assert(start !== undefined, 'expected a start position when splicing')
-    jumps.unshift([start, start + slice.length - 1])
+    jumps.push([start, start + slice.length - 1])
```
remcohaszing, line 249: It looks like jumps is an array of `[number, number]` tuples, which always ends up having the same length as `breaks.length`.

If this is such a hot path, maybe we can optimize it further by:

- Making `jumps` an array of fixed length instead of pushing.
- Splitting `jumps` into 2 arrays of `number[]`.
- Turning those two arrays into arrays of type `UintNArray`, where I don’t know the appropriate value of `N`.

wooorm: Yes! But I’d personally really want to look at the edit map data structure which I used in Rust: https://github.com/wooorm/markdown-rs/blob/60db8e5896be05d23137a6bdb806e63519171f9e/src/util/edit_map.rs#L1-L9.

wooorm:
Thanks, [released](https://github.com/micromark/micromark/releases/tag/micromark-util-subtokenize%402.0.1)!

Issue #23, opened 2020-09-29
Performance improvement: linked lists for events


wooorm:
### Subject of the feature

Given that on large markdown files we are dealing with tons (literally, 100k or so) of events, improving performance might be switching from arrays to linked event objects.

### Problem

Operations on big arrays can be slow, such as https://github.com/micromark/micromark/pull/21.
Switching to linked lists adds complexity (while removing it in certain other cases!), but will *probably/hopefully* improve perf.

### Alternatives

We’re already using really fast array methods. And everything is mutating already. Maybe linked lists won’t net a lot.

3y3:
Things about linked list
https://gist.github.com/3y3/cd4d9272a5a309502ae7c2bf961788d1

This improves speed x2-x3 on [this](https://github.com/wooorm/a-dump-of-markdown) benchmark, but slows down processing of small files in markdown-it [benchmark](https://github.com/markdown-it/markdown-it/tree/master/benchmark)

wooorm:
Ah, cool! Thanks for checking and providing code :)

PR #171, opened 2024-03-28
Refactor to improve postprocess performance

robsimmons:
### Initial checklist

### Description of changes

This addresses #23 and more-or-less solves #169 by making the event-holding data structure a double-array structure, a `SpliceBuffer` (effectively just a [gap buffer](https://en.wikipedia.org/wiki/Gap_buffer)) that allows efficient repeated splices. On the benchmark from #169, the largest samples go from taking about 55 seconds to parse to taking less than 2 seconds, a better-than-25x speedup. (And an asymptotic speedup, as I believe that family of examples goes from being approximately quadratic to approximately linear.)

Two things worth noting:

1. I think this is taking a different direction than the edit map @wooorm pointed to [here](https://github.com/wooorm/markdown-rs/blob/60db8e5896be05d23137a6bdb806e63519171f9e/src/util/edit_map.rs#L1-L9) in the discussion of #170.
2. There are a lot of other array data structures in micromark that would benefit from using the SpliceBuffer - if a splicebuffer is just being used like a regular array, it has very little overhead, and using a SpliceBuffer instead of the array would have made pr #170 unnecessary, and SpliceBuffer could be used to get rid of every place that the `splice` helper from `micromark-util-chunked` is being used. I kept the usage of SpliceBuffer here very narrow because any more pervasive change would have changed exposed types. But would a wider refactor be valuable? 

Assuming this seems like a reasonable edit, should the splicebuffer data structure live in a different util package? (pacakges/micromark-util-splice-buffer)?


robsimmons:
A note: I initially used proxies so that the array accesses `event[i]` wouldn't need to be changed to `event.get(i)`, but based on the markdown-it performance benchmarks I got a small performance improvement by not adding a proxy.

wooorm:
Heya! Cool!

Have you looked at the edit map idea?

The reason I haven’t implemented it here yet, is the same as why you haven’t implemented your direction everywhere too: breaking changes across the system.

I don’t particularly care more for one or the other. Just, the code needs to move towards alignment in both languages.

robsimmons:
I'm happy to look more into the edit map idea, but I've only written a bit of rust (and, fatally, have read less than I've written), so I didn't immediately understand what was going on with edit maps. I _do_ think that this data structure has something of an advantage in clarity if you've found yourself reversing a lot in the Rust implementation - a gap buffer is gonna give you more "it's easier to understand what the code is doing without all these reverses" at the cost of "it's harder to see what the expensive operations are"

Here's a few concrete-ish question to start:
 - _How_, at a high level, is the edit map making the process more efficient?
 - The edit map is a `Vec<(usize, usize, Vec<Event>)>` in edit_map.rs, which doesn't really give me much insight into how the edit map relates to a vector of events. What is the vector a vector of/what are the usizes/what's the child vector, and how does each one related to the underlying event vector being manipulated?

robsimmons:
Okay, I think I see it - at a high level, all the splices have quadratic behavior for the same reason that it's quadratic to get the concatenation of a lists of lists by appending in a loop. The edit map is kind of batching all those updates in the same way that doing a batch-append is.

Things I'd like to do (or would like someone else to do but I don't think that's likely):
- Confirm that the rust implementation doesn't have the performance issues I described in #169 (sanity check that using edit maps would actually fix the problems I encountered in the JS implementation)
- Write up some additional high-level documentation of this parsing algorithm to make sure I understand it, seems like that would be valuable and making explanatory things is kind of my jam
- Apply edit maps to the JS implementation

That's... several steps! I'll have to think about whether I have capacity for that big of a project next week when I'm back from Eclipse Related Travel. (I'm assuming there's probably not budget for a bounty on this or anything via opencollective, would love to be told I'm wrong.)

In the meantime I wonder where that leaves this PR - I'd argue this isn't really _increasing_ divergence from the Rust implementation, and if splice buffers were treated as an external dependency this would only be something like a +86 -13 change. That said, it's _not_ an external dependency and micromark doesn't have external dependencies, and I'm not gonna push hard for this particular change (that would be [big Jia Tan energy](https://mastodon.gamedev.place/@zarfeblong/112202381516546376)). At the same time, it _would_ dramatically improve my Astro build times, which would be nice - but I've decreased MDX file size enough that I'm not getting 60 minute build times (99% parsing MDX) anymore, so I don't have any fire alarms going off in my house, personally.

Murderlon:
> That's... several steps! I'll have to think about whether I have capacity for that big of a project next week when I'm back from Eclipse Related Travel. (I'm assuming there's probably not budget for a bounty on this or anything via opencollective, would love to be told I'm wrong.)

Just chipping in to say that from my perspective this is very valuable work and it requires expertise that few contributors have. So in my book this would be an elegible case for funds. I don't know how much and it needs @wooorm's approval of course, but those are my thoughts.

Taking one step back (without having deep knowledge on this myself) this PR seems to practically solve the problem, seems high quality, comes with tests. and well written code comments. Generally speaking it would take a very strong case to not merge this and thus prevent the ecosystem from enjoying this speedup.

I don't have the expertise to push hard for this or the edit maps idea, but I'd just want to say that it seems like a shame to let this go unless the edit maps idea is crystal clear. Perhaps a middle ground is also possible, merge and release this and patiently explore other ideas.

wooorm:
> So in my book this would be an elegible case for funds.

I mostly agree, indeed.
Though, there’s a lot of work to be done that is eligible to funds. There are few funds. There are 218+ open issues. Where do funds have the most use? As you correctly point out: this PR already solves a lot of the issue.

I made the algorithm in JS, then improved it in Rust. I backported most but not all. It can be assumed that both a) the 2nd Rust version is better than the 1st JS version; b) there are more improvements to be made in the Rust version

The best case scenario is for *both* to improve.
The worst case scenario is for 2 forks to appear.

Another bad case is work that will be discarded. So yeah, I don’t think this PR should be discarded, but it’s also why I very quickly linked the edit maps & rust version and want to focus on that.

> That's... several steps! I'll have to think about whether I have capacity for that big of a project next week 

IMO, what we should do, depends a lot on what you want Rob! Doesn’t have to be next week. We do indeed have some funds.

Rust is fun. Having the same algorithm in Rust *and* in JS is also a good way to learn.
Markdown is really messy but that’s also kinda fun and very practically useful to people: as you notice, your work directly impacts a ton of folks.
Looks like you have a CS background; that’s I don’t have. But this is still parsers and tokenizers and ASTs and such. So I’m sure there’s a lot of CS ideas that you could apply to improve things.

remcohaszing:
Just adding that I fully agree with @Murderlon. Even if this diverges from the current Rust implementation, the JS ecosystem would really benefit from the speedups.

wooorm (approved):

wooorm:
Thanks, [released](https://github.com/micromark/micromark/releases/tag/micromark-util-subtokenize%402.0.1)!

PR #172, opened 2024-04-22
Fix email autolink starting w/ `@`

yannickgloster:
### Initial checklist

### Description of changes

The CommonMark spec defines an [email autolink](https://spec.commonmark.org/0.31.2/#email-autolink) as an email that follows [non-normative regex from the HTML5 spec](https://html.spec.whatwg.org/multipage/forms.html#e-mail-state-(type=email)) encased in `< >`.

This means that `<@example.com>` should not be autolinked as `@example.com` is not a valid email address.

https://spec.commonmark.org/dingus/?text=%3C%40example.com%3E%0A


wooorm:
Sweet, thanks!

PR #175, opened 2024-06-01
feat: fix imports tree-shake; micromask/index must not contain funcs,…

denisx:
### Initial checklist

*   [х] I read the support docs 
*   [х] I read the contributing guide 
*   [х] I agree to follow the code of conduct 
*   [х] I searched issues and couldn’t find anything (or linked relevant results below) 
*   [х] If applicable, I’ve added docs and tests

### Description of changes

Fixing imports for tree-shake: micromask/index must not contain funcs. Make separate file for it.


denisx:
need help with last ci-test 🙄

ChristianMurphy:
❓
Please provide more information, it's unclear what this PR is trying to solve.

denisx:
Main idea: if i choose this `import {compile} from 'micromark` in my code, i get whole bundle, because `micromark()` function use all imports of index file. And tree-shake dont't work. The solving is separate function to file, and make `index.js` only re-exports

ChristianMurphy:
This module is side effect free, and is marked as such
https://github.com/micromark/micromark/blob/26c38634092b164f9d43705f61265c1c7c576dad/packages/micromark/package.json#L42

That should be able to be tree shaken.
This feels more like a bug in your bundler/tree-shaker than in micromark.

wooorm:
> if i choose this import {compile} from 'micromark in my code, i get whole bundle [...] The solving is separate function to file

It sounds like you use a buggy tree shaker. Which one do you use? Why do you not raise an issue there?

PR #176, opened 2024-06-16
Fix `micromark-build` in `dev` folders

puqeko:
### Initial checklist

### Description of changes

**Problem:** Cloning micromark (main branch) into `~/dev/` (or any path containing `/dev/`) and running `npm install && npm run build` fails with `Cannot find module...`.

**Cause:** micromark-build replaces the first occurrence of `/dev/` in the full file path instead of the last occurrence when moving `.js` files up a directory.

**Proposed Solution:** This PR adds a negative lookahead to the regular expression in `.replace` so that the **last** occurrence of `/dev/` is replaced correctly.

### DIFF
--- a/packages/micromark-build/index.js
+++ b/packages/micromark-build/index.js
@@ -19,7 +19,7 @@ let index = -1
 
 while (++index < files.length) {
   const input = new URL(files[index], root + '/')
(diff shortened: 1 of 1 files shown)


wooorm:
Thanks! :)

PR #185, opened 2024-11-10
Refactor to improve resolveAllLabelEnd performance

PerfectPan:
### Initial checklist

* [x] I read the support docs 
* [x] I searched issues and couldn’t find anything (or linked relevant results below) 
* [x] If applicable, I’ve added docs and tests

### Description of changes

I use the Unified ecosystem to compile Markdown strings in my service. Recently, I encountered a serious performance issue when the input string contains a large number of '[]' patterns. After profiling the problematic input, I found that the majority of time consumption occurs in this function: https://github.com/micromark/micromark/blob/main/packages/micromark-core-commonmark/dev/lib/label-end.js#L40-L59.

The issue arises because the input string contains many '[]' sequences, which triggers repeated array splicing in the events array, leading to significant performance overhead.

To address this, I attempted to optimize the function by removing the use of splice and instead creating a new array. While this approach might be slower in scenarios where the main logic is not heavily utilized, it significantly reduces compilation time when processing Markdown strings with numerous '[]' patterns.

You can see my research and results here: https://github.com/PerfectPan/micromark-label-end-performance. With this optimization, the processing time for a Markdown string containing 10,000 instances of '[]' was reduced from 500ms to just 3ms.

### DIFF
--- a/packages/micromark-core-commonmark/dev/lib/label-end.js
+++ b/packages/micromark-core-commonmark/dev/lib/label-end.js
@@ -43,11 +43,7 @@ function resolveAllLabelEnd(events) {
   const newEvents = []
   while (++index < events.length) {
     const token = events[index][1]
-    if (newEvents.length > 0) {
-      push(newEvents, [events[index]])
-    } else {
-      newEvents.push(events[index])
-    }
+    newEvents.push(events[index])
 
     if (
       token.type === types.labelImage ||
@@ -57,15 +53,15 @@ function resolveAllLabelEnd(events) {
       // Remove the marker.
       const offset = token.type === types.labelImage ? 4 : 2
       token.type = types.data
-      // Exit effect can skip the check
-      index += offset + 1
-      if (index < events.length) {
-        push(newEvents, [events[index]])
-      }
+      index += offset
     }
   }
 
-  splice(events, 0, events.length, newEvents)
+  // If the events are equal, we don't have to copy newEvents to events
+  if (events.length !== newEvents.length) {
+    splice(events, 0, events.length, newEvents)
+  }
+
   return events
 }
 
--- a/packages/micromark-core-commonmark/dev/lib/label-end.js
+++ b/packages/micromark-core-commonmark/dev/lib/label-end.js
@@ -39,9 +39,15 @@ const referenceCollapsedConstruct = {tokenize: tokenizeReferenceCollapsed}
 /** @type {Resolver} */
 function resolveAllLabelEnd(events) {
   let index = -1
-
+  /** @type {Array<Event>} */
+  const newEvents = []
   while (++index < events.length) {
     const token = events[index][1]
+    if (newEvents.length > 0) {
+      push(newEvents, [events[index]])
+    } else {
+      newEvents.push(events[index])
+    }
 
     if (
       token.type === types.labelImage ||
@@ -49,12 +55,17 @@ function resolveAllLabelEnd(events) {
       token.type === types.labelEnd
     ) {
       // Remove the marker.
-      events.splice(index + 1, token.type === types.labelImage ? 4 : 2)
+      const offset = token.type === types.labelImage ? 4 : 2
       token.type = types.data
-      index++
+      // Exit effect can skip the check
+      index += offset + 1
+      if (index < events.length) {
+        push(newEvents, [events[index]])
+      }
     }
   }
 
+  splice(events, 0, events.length, newEvents)
   return events
 }
 


packages/micromark-core-commonmark/dev/lib/label-end.js
```diff
@@ -39,22 +39,33 @@ const referenceCollapsedConstruct = {tokenize: tokenizeReferenceCollapsed}
...
-
+  /** @type {Array<Event>} */
+  const newEvents = []
   while (++index < events.length) {
     const token = events[index][1]
+    if (newEvents.length > 0) {
+      push(newEvents, [events[index]])
+    } else {
+      newEvents.push(events[index])
+    }
```
wooorm, line 50: This looks rather weird? Why not:

```suggestion
    push(newEvents, [events[index]])
```

Or:


```suggestion
    newEvents.push(events[index])
```

packages/micromark-core-commonmark/dev/lib/label-end.js
```diff
@@ -39,22 +39,33 @@ const referenceCollapsedConstruct = {tokenize: tokenizeReferenceCollapsed}
...
       token.type === types.labelEnd
     ) {
       // Remove the marker.
-      events.splice(index + 1, token.type === types.labelImage ? 4 : 2)
+      const offset = token.type === types.labelImage ? 4 : 2
       token.type = types.data
-      index++
+      // Exit effect can skip the check
+      index += offset + 1
+      if (index < events.length) {
```
wooorm, line 62: This looks potentially weird? How often is `index < events.length`, `false`?

This check is also already performed by the loop itself? Looking quickly, it looks like `if (index < events.length) { push(newEvents, [events[index]]) }` can be dropped entirely because the loop would do the same anyway?

wooorm (commented):
Hi!

The code look a bit weird to me, like it can be improved:

packages/micromark-core-commonmark/dev/lib/label-end.js
PerfectPan: Yeah, because push function will not push events when the array is empty, I don't know the correct way to use it. So I wrote down the strange code. Do you have any advise?

https://github.com/micromark/micromark/blob/main/packages/micromark-util-chunked/dev/index.js#L82-L89

CleanShot 2024-11-11 at 21 56 22@2x

PerfectPan: I think the `newEvents.push(events[index])` is ok，do you recommend use this way if the push function need non-empty array.

wooorm: How it works is explained in the comment.

You now wrote code that always creates arrays with one item for each item. That is potentially millions of arrays. That is not a good idea.

This specific code is for other cases, where it can be fast.

If your use case is to always only ever push one item — this is not the function for you. You can use `Array#push`.

The solution *with* the specific custom push is `newEvents = push(newEvents, [events[index]])`. But, this custom push is not good for this use case. You should be able to use the regular push: `newEvents.push(events[index])`.

packages/micromark-core-commonmark/dev/lib/label-end.js
PerfectPan: Beacuse the logic is when skip some events, we need to skip the exit effect as well. So the previous code will increment the index after splice the events. 

If we want to create the new effects we need to push it into newEvents to keep the same logic.


I think the code may a little werid as well. Let me try another implemention.

packages/micromark-core-commonmark/dev/lib/label-end.js
PerfectPan: OK, I will try another implementation.

packages/micromark-core-commonmark/dev/lib/label-end.js
PerfectPan: > Beacuse the logic is when skipping some events, we need to skip the exit effect as well. So the previous code will increment the index after splice the events.
> 
> If we want to create the new effects we need to push it into newEvents to keep the same logic.
> 
> I think the code may a little werid as well. Let me try another implementaion.

PerfectPan: @wooorm Thx for the patience reviewing. I have done some code refactoring based on your previous code review. Could you please help review the updated code when you have time?

wooorm (approved):

wooorm:
This is great, thanks!

wooorm:
Released in [`4.0.1`](https://github.com/micromark/micromark/releases/tag/4.0.1)!

PR #196, opened 2025-03-30
Fix type error of Stream.Writable.write

tats-u:
### Initial checklist

* [x] I read the support docs 

### Description of changes

If you remove `node_modules` and re-run `npm install`, `npm run build` will fail due to the fact that the signature of `Stream.Writable.write` has been changed and that change caused a type error of TypeScript.

https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback

This change passes the default value `"utf8"` to supress the type error.

### DIFF
--- a/test/util/slow-stream.js
+++ b/test/util/slow-stream.js
@@ -20,7 +20,7 @@ export function slowStream(value, encoding) {
     if (index === value.length) {
       stream.end()
     } else {
-      stream.write(value.slice(index, ++index), encoding ?? 'utf8', tick)
+      stream.write(value.slice(index, ++index), encoding || 'utf8', tick)
     }
   }
 
--- a/test/util/slow-stream.js
+++ b/test/util/slow-stream.js
@@ -20,7 +20,7 @@ export function slowStream(value, encoding) {
     if (index === value.length) {
       stream.end()
     } else {
-      stream.write(value.slice(index, ++index), encoding, tick)
+      stream.write(value.slice(index, ++index), encoding ?? 'utf8', tick)
     }
   }
 


tats-u:
I believe this `slowStream` is not used by react-markdown and this PR is safe to be merged.

test/util/slow-stream.js
```diff
@@ -20,7 +20,7 @@ export function slowStream(value, encoding) {
     if (index === value.length) {
       stream.end()
     } else {
-      stream.write(value.slice(index, ++index), encoding, tick)
+      stream.write(value.slice(index, ++index), encoding ?? 'utf8', tick)
```
wooorm, line 23:
```suggestion
      stream.write(value.slice(index, ++index), encoding || 'utf8', tick)
```

I guess a regular `||` should be fine?

I like sticking with things that are more often available: more authors know `||`, more engines support it too, it should be equivalent here, using the latest features has no real benefit

As for the `react-markdown` error: I guess that *it* fails now, on its own. Probably because `react` recently released 19.1

tats-u: `||` also works here, but you should note:

- `??` was invented in C# 20 years ago (2005), and adopted by Perl (2007), Dart (2015) Swift (2015), and PHP (2018) before adopted by ES2020. It is not the property of JS alone.
- `??` has been available in JS for 5 years, and it is supported by all Node.js supporting ESM and [all practical versions of browsers](https://caniuse.com/mdn-javascript_operators_nullish_coalescing). (e.g. iOS 13.4+, c.f. WebP images is supported in [iOS 14+](https://caniuse.com/?search=webp))
- Using the `||` operator instead may mislead a reader who does not know the type details of BufferEncoding into thinking that it could also take empty characters.

However, the final decision is up to you. If you still prefer `||`, I will change this line.

wooorm: Yes, I prefer `||`. When in JavaScript, do as JavaScripters do. I use things that always work, always. Only when special things are needed, do I use special things. It is not needed to act like the C compiler or act as the most recent JavaScript engine. Thank you!

wooorm (approved):

wooorm:
Thanks!

PR #210, opened 2025-12-07
Allow sms: and tel: protocols in href

TownCube:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Allow `sms:` and `tel:` protocols in `href` links.

These are safe in the same way as `mailto:` because they invoke either the messaging/dialer apps with the number prefilled but don't send/call unless the user confirms, similar to an email message with the address prefilled by `mailto:`

The code uses GitHub as a justification for the existing safe protocols which does not include `sms:/tel:`. However, [WordPress deems them safe](https://developer.wordpress.org/reference/functions/wp_allowed_protocols/) as part of its URL sanitising.


wooorm:
Hi! The idea is to stick with GH, which is safe and a common use case, to not have to deal with such questions.
You can turn off this behavior and use your own sanitizer to use other things! That is also what using mdast/remark -> hast/rehype, with specifically `hast-util-sanitize`/`rehype-sanitize` allows!

PR #218, opened 2026-05-02
Replace `debug` with `obug`

ocavue:
### Initial checklist

* [x] I read the support docs 

### Description of changes

`micromark` currently has `debug` as one of its dependencies. `debug` is a CommonJS package, and it doesn't work well in non-Node.js environments like browsers and Cloudflare workers.

I suggested that we replace `debug` with [`obug`](https://npmx.dev/package/obug), a pure ESM fork of `debug`. `obug` is already used in many popular projects, including [vite](https://github.com/vitejs/vite/pull/21107).


remcohaszing:
I like it, but I see no documentation on the `obug` support window. `micromark` supports Node.js 16+. `obug` states it supports modern Node.js versions, but it doesn’t specify anywhere what that means.

ocavue:
> I see no documentation on the `obug` support window.


`obug` does support Node.js on Windows as it simply uses `process.env`. There are also Windows-based CI workflows in `obug`'s repo.

I copied and tweaked the README from the original `debug`, and in this PR: https://github.com/sxzz/obug/pull/2, I hope it makes it clearer.


> `obug` states it supports modern Node.js versions, but it doesn’t specify anywhere what that means.

`obug` only uses the following Node.js API, as shown [here](https://app.unpkg.com/obug@2.1.1/files/dist/node.js#L2-3). All of them are available since Node.js v10.

```
import { isatty } from "node:tty";  
import { formatWithOptions, inspect } from "node:util";
```

As for modern Node.js, since `obug` is pure ESM, it requires Node.js to support ESM. Since `micromark` is also pure ESM, I don't think this is a blocker.

ChristianMurphy:
Node 22 is min for obug

https://github.com/sxzz/obug/blob/e39bee9fcf1e6739323eede6cd53f3276b9eed5e/.github/workflows/unit-test.yml#L13

https://github.com/sxzz/workflows/blob/fec52358a3ac0d8d03df5f4ae67ad7e4c8a91137/.github/workflows/unit-test.yml#L19

ocavue:
`obug` only runs CI on Node.js v22+ doesn't mean it only supports Node.js v22+.

ChristianMurphy:
It does. There is a difference between intentionally and accidentally/incidentally supported.
It intentionally supports and tests 22+.
It may accidentally/incidentally not use any features from lower node versions, but unless tested/promised, that isn't a support we can rely on.

Murderlon:
Seems like it's not going to happen due to support window.

ChristianMurphy:
It could happen in the next major, when we drop old/unsupported node runtimes.
But I don't have a timeline on when the next major would be coming.

PR #223, opened 2026-05-20
Drop `debug` dependency

fuma-nama:
### Initial checklist

* [x] I read the support docs 

### Description of changes

Replace `debug` with a runtime `console.log()` + babel plugin, this is by possible using existing plugins with dead code elimination:

```ts
if (DEBUG) {
  console.log("will be removed by dead code elimination when combined with define plugin")
}
```

for simplicity, but let me know if you have other preferences. This allows usage of `development` export under browser environment (e.g. in dev server), without bundling it (thanks to ESM)

### DIFF
--- a/packages/micromark-build/index.js
+++ b/packages/micromark-build/index.js
@@ -17,11 +17,16 @@ import {resolve} from 'import-meta-resolve'
 import RelativizeUrl from 'relativize-url'
 
 const root = pathToFileURL(process.cwd())
+const debugConstantPath = pathToFileURL(
+  path.join(process.cwd(), 'dev/debug.js')
+)
 const files = await glob('./dev/**/*{.d.ts.map,.d.ts,.js}', {cwd: root})
 let index = -1
 
 while (++index < files.length) {
   const input = new URL(files[index], root + '/')
+  if (input.href === debugConstantPath.href) continue
+
   const output = new URL(input.href.replace(/\/dev\/(?!.*\/dev\/)/, '/'))
   const inputRelative = RelativizeUrl.relativize(input, root)
   const outputRelative = RelativizeUrl.relativize(output, root)
@@ -54,7 +59,13 @@ while (++index < files.length) {
     })
     .filter(Boolean)
 
-  const result = await babel(String(await fs.readFile(input)), {
+  // Check if ./debug.js file exists before inlining
+  try {
+    await fs.access(debugConstantPath)
+    modules.push(debugConstantPath.href)
+  } catch {}
+
+  const result = await babel(await fs.readFile(input, 'utf-8'), {
     filename: fileURLToPath(input),
     plugins: [
       [
@@ -70,8 +81,8 @@ while (++index < files.length) {
           ]
         }
       ],
-      removeDebugLog,
-      ['babel-plugin-inline-constants', {modules}]
+      ['babel-plugin-inline-constants', {modules}],
+      'babel-plugin-minify-dead-code-elimination'
     ]
   })
 
@@ -82,38 +93,4 @@ while (++index < files.length) {
   console.log('%s (from `%s`)', outputRelative, inputRelative)
 
   await fs.writeFile(output, result.code)
-}
-
-/**
- * @returns {PluginObj}
- */
-function removeDebugLog() {
-  const name = '$logDebug'
-
-  return {
-    name: 'remove-logdebug',
-    visitor: {
-      CallExpression(path) {
-        const callee = path.node.callee
-
-        // Remove `$logDebug()`.
-        if (t.isIdentifier(callee, {name})) {
-          path.remove()
-        } else if (
-          t.isMemberExpression(callee) &&
-          t.isIdentifier(callee.property, {name})
-        ) {
-          path.remove()
-        }
-      },
-      VariableDeclaration(path) {
-        if (
-          path.node.declarations.length === 1 &&
-          t.isIdentifier(path.node.declarations[0].id, {name})
-        ) {
-          path.remove()
-        }
-      }
-    }
-  }
 }
--- a/packages/micromark-build/index.js
+++ b/packages/micromark-build/index.js
@@ -90,24 +90,26 @@ while (++index < files.length) {
  *   Babel plugin.
  */
 function removeDebugLog({types: t}) {
+  const name = '$logDebug'
+
   return {
     name: 'remove-logdebug',
     visitor: {
       CallExpression(path) {
         const callee = path.node.callee
 
-        // Remove `logDebug()`.
-        if (t.isIdentifier(callee, {name: 'logDebug'})) {
+        // Remove `$logDebug()`.
+        if (t.isIdentifier(callee, {name})) {
           path.remove()
         } else if (
           t.isMemberExpression(callee) &&
-          t.isIdentifier(callee.property, {name: 'logDebug'})
+          t.isIdentifier(callee.property, {name})
         ) {
           path.remove()
         }
       },
       FunctionDeclaration(path) {
-        if (t.isIdentifier(path.node.id, {name: 'logDebug'})) {
+        if (t.isIdentifier(path.node.id, {name})) {
           path.remove()
         }
       }
--- a/packages/micromark-build/index.js
+++ b/packages/micromark-build/index.js
@@ -106,8 +106,11 @@ function removeDebugLog() {
           path.remove()
         }
       },
-      FunctionDeclaration(path) {
-        if (t.isIdentifier(path.node.id, {name})) {
+      VariableDeclaration(path) {
+        if (
+          path.node.declarations.length === 1 &&
+          t.isIdentifier(path.node.declarations[0].id, {name})
+        ) {
           path.remove()
         }
       }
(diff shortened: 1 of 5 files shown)


packages/micromark/dev/lib/create-tokenizer.js
```diff
@@ -715,3 +712,8 @@ function serializeChunks(chunks, expandTabs) {
 
   return result.join('')
 }
+
+/** @param {string} v  */
+function $logDebug(v) {
+  console.log(`[micromark] ${v}`)
```
remcohaszing: The `debug` package respects the `DEBUG` environment variable. This implementation does not. That would be a breaking change.

Tyical console methods accept a number of arguments to log. These are typically separated by a space in the output. IMO it’s nicer to pass multiple arguments instead of concatenating strings.

```js
console.log('[micromark]', v)
```

We can take this a step further. Some environments (e.g. Chrome DevTools) log the location where console methods are called. In this case, that would be `create-tokenizer.js:718`

If we remove the wrapper method, and call the console method directly, that would preserve the positional information. We can achieve this and still customize the behaviour using `Function.bind()`.

```js
const $logDebug = console.log.bind(console, '[micromark]')
```

We can take this one step further and more fitting log level.

```js
const $logDebug = console.debug.bind(console, '[micromark]')
```

Maybe this should be the browser implementation. The default implementation could respect the `DEBUG` environment variable. This can be configured using package imports.

```js
export const $logDebug = process.env.DEBUG?.includes('micromark')
  ? console.debug.bind(console, '[micromark]')
  : () => {}
```

packages/micromark-build/index.js
```diff
@@ -80,3 +80,39 @@ while (++index < files.length) {
 
   await fs.writeFile(output, result.code)
 }
+
+/**
+ *  @import {PluginObj} from '@babel/core'
```
remcohaszing: Type imports should go in their own dedicated JSDoc comment on top of the module (above the imports, below the shebang).

packages/micromark-build/index.js
```diff
@@ -80,3 +80,39 @@ while (++index < files.length) {
...
+ *
+ * @param {{types: typeof import('@babel/core').types}} api
```
remcohaszing: This import can be done with an `@import` tag. But also, doesn’t Babel provide an actual type for the plugin API?

remcohaszing (changes requested):
I have some suggestions. But also, I don’t think of dropping any CJS dependencies as a goal of micromark.

I would love to get some input on this from other maintainers who are more involved with micromark internals.

Murderlon:
> I have some suggestions. But also, I don’t think of dropping any CJS dependencies as a goal of micromark.
> 
> I would love to get some input on this from other maintainers who are more involved with micromark internals.

To start off, I am not someone involved with micromark internals. However my two cents: in the age of supply chain attacks, having less dependencies is a better idea than ever, especially if the code is minimal and rarely/never changes. It's better to own things than to have dependencies (when feasible).

packages/micromark-build/index.js
fuma-nama: cannot remember the details as I haven't used babel for a long time, I only found:

```ts
export type PluginTarget = string | object | ((...args: any[]) => any);
```

in the types package, I might have overlooked something I guess.

packages/micromark/dev/lib/create-tokenizer.js
fuma-nama: hmm I see, will change that. But `process.env` is a Node.js thing, probably need some extra checks to work on browsers.

fuma-nama:
yep it's one of my motivations, and one of the advantages of ESM is that it works on browsers without additional bundling or transpiling. Currently to use the npm version (not `https://esm.sh` etc) on browser, it needs something like Vite + optimize deps to transpile the cjs deps, but `micromark` is already ESM-only, removing the last cjs dep allows it to take effect, which is great I think.

packages/micromark/dev/lib/create-tokenizer.js
```diff
@@ -715,3 +712,8 @@ function serializeChunks(chunks, expandTabs) {
 
   return result.join('')
 }
+
+const $logDebug =
+  typeof process !== 'undefined' && process.env.DEBUG?.includes('micromark')
```
ChristianMurphy: Bug. The `n/prefer-global/process` lint rule fails on this line because `process.env.DEBUG` is read without `import process from 'node:process'` at the top of the file. Running `xo` locally on the PR head produces two errors at col 10 and col 37. The `build/index.js` change in this PR already adds the same import for itself; adding it here clears the lint.

ChristianMurphy: Question. The substring check diverges from the `debug` package's pattern semantics in two places. `DEBUG=*` (the common \"enable everything\" pattern) returns false because the literal `micromark` is not in `*`. And `DEBUG=somethingmicromark` matches because it is a bare substring. If the goal is a minimal maintainer-only flag, the current shape is fine; if existing `DEBUG` patterns developers already have set should keep working for micromark too, splitting on whitespace and comma and matching each entry against `micromark` or `*` covers both cases in one extra line. See the [`debug` README on wildcards](https://github.com/debug-js/debug#wildcards).

packages/micromark-build/index.js
```diff
@@ -80,3 +83,37 @@ while (++index < files.length) {
...
+        // Remove `$logDebug()`.
+        if (t.isIdentifier(callee, {name})) {
+          path.remove()
+        } else if (
+          t.isMemberExpression(callee) &&
+          t.isIdentifier(callee.property, {name})
+        ) {
+          path.remove()
+        }
+      },
```
ChristianMurphy: Bug. `path.remove()` on a `CallExpression` produces the desired statement strip only when the call sits at statement position. When the call appears inside an expression, babel removes the call node and silently rewrites the surrounding expression. I ran the plugin on three shapes:
 
- `const x = $logDebug('a') || 'b'` becomes `const x = 'b'`
- `console.log($logDebug('a'))` becomes `console.log()`
- `return $logDebug('a')` becomes `return;`
 
The current dev source only uses `$logDebug` as a top-level statement, so today's build is correct, but the plugin gives no signal if a future caller uses it in expression position. Two options: guard with `path.parentPath.isExpressionStatement()` before removing, or replace with `t.identifier('undefined')` so misuse stays visible. See [babel-traverse](https://babeljs.io/docs/babel-traverse) for the path API.

packages/micromark-build/index.js
```diff
@@ -80,3 +83,37 @@ while (++index < files.length) {
...
+      VariableDeclaration(path) {
+        if (
+          path.node.declarations.length === 1 &&
+          t.isIdentifier(path.node.declarations[0].id, {name})
+        ) {
+          path.remove()
+        }
+      }
```
ChristianMurphy: Suggestion. The visitor only matches declarations with exactly one declarator. If anyone later writes `const $logDebug = ..., other = 1`, the helper stays in the production bundle. Targeting the matching `VariableDeclarator` instead avoids the gap. The current declaration is a single dedicated `const`, so this is theoretical.

packages/micromark/dev/lib/create-tokenizer.js
fuma-nama: Hmm, to clarify, my original goal wasn't to re-implement the same API surface like debug do, I think I can replicate this for backward compat, but I am neither sure if the flag was intentional for maintainer-only use.

packages/micromark-build/index.js
fuma-nama: I was thinking if we can simply do `if (IS_DEBUG)` + a babel plugin, would agree the current solution isn't good, it was simply built minimally based on the current usage (so far it's the only file that actually uses `debug`).

fuma-nama:
I just realized there is already a define plugin in the build pipeline (wooorm called it inline constants), it's possible to use traditional approaches like `if (IS_DEBUG)` without much hacking with babel plugins.

The new code works pretty well with babel's official dead code elimination plugin, and should fully align with `debug`'s behaviour now.

hushino:
Is this related to the spam of prints in terminal that can cause if run dev in tanstack start during ssr?  
for reference i fix it  with this code in vite config that map to production entry points: 
resolve: {
    tsconfigPaths: true,
    alias: {
      'micromark/stream': require.resolve('micromark/stream'),
      'micromark': require.resolve('micromark'),
    },
  },

ChristianMurphy:
> Is this related to the spam of prints in terminal that can cause if run dev in tanstack start during ssr?

No, this PR is a dependency update that keeps the logging behavior largely as is.
The spam of logs you're seeing is likely because tanstack is overbroad in where it gets develop scoping of dependencies.
The default build, which tanstack likely should use, doesn't even have the logging code in the source https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/packages/micromark/package.json#L55-L62

wooorm:
I think it is better to spend your energy on improving `debug` (and Vite). Or having another tool like it that does work and is also popular.

micromark is really complex and debugging is very importing. But maintaining a debugger here is... not a goal?

PR #224, opened 2026-05-25
fix: allow hard breaks when tabs precede trailing spaces

### LINKED ISSUE #206: Trailing tab "cancels" hard line break
### Initial checklist

- [x] I read the [support docs](https://github.com/micromark/.github/blob/main/support.md)
- [x] I read the [contributing guide](https://github.com/micromark/.github/blob/main/contributing.md)
- [x] I agree to follow the [code of conduct](https://github.com/micromark/.github/blob/main/code-of-conduct.md)
- [x] I searched [issues](https://github.com/search?q=user%3Amicromark&type=issues) and [discussions](https://github.com/orgs/micromark/discussions) and couldn’t find anything (or linked relevant results below)

### Affected package

micromark

### Steps to reproduce

```ts
import { fromMarkdown } from 'mdast-util-from-markdown'
import { inspect } from 'unist-util-inspect'
console.log(
  inspect(
    fromMarkdown(`foo	  
bar`)
  )
)
console.log(
  inspect(
    fromMarkdown(`foo  
bar`)
  )
)
```

### Actual behavior

```
root[1] (1:1-2:4, 0-10)
└─0 paragraph[1] (1:1-2:4, 0-10)
    └─0 text "foo\nbar" (1:1-2:4, 0-10)
root[1] (1:1-2:4, 0-9)
└─0 paragraph[3] (1:1-2:4, 0-9)
    ├─0 text "foo" (1:1-1:4, 0-3)
    ├─1 break (1:4-2:1, 3-6)
    └─2 text "bar" (2:1-2:4, 6-9)
```

### Expected behavior

Both trees have line breaks. The [spec](https://spec.commonmark.org/0.31.2/#hard-line-break) doesn't seem to tell that tab should cancel line break. 
Users often bring trailing spaces and tabs when typing text in WYSIWYG editors

### Runtime

_No response_

### Package manager

_No response_

### Operating system

_No response_

### Build and bundle tools

_No response_

wyf027:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This fixes hard line break detection when tabs appear before trailing spaces, such as `foo\t  \nbar`.

Hard breaks require two or more spaces immediately before the line ending. Tabs earlier in the line suffix are still trimmed as line suffix whitespace, but they no longer cancel a hard break when the final contiguous run is at least two spaces. Tabs directly before the line ending, such as `aaa  \t\nbb`, still produce soft breaks.

Test plan:

* [x] Updated mixed suffix tests in `test/io/text/hard-break.js`
* [x] Added regression test for #206
* [x] `npm run build`
* [x] `npm run format`
* [x] `npm run test-api-dev` (1928 tests)

Closes #206


ChristianMurphy (changes requested):
Thanks for digging into this! I applied the patch to a copy of `micromark@4.0.2` and ran it against [commonmark.js](https://github.com/commonmark/commonmark.js): all 6 cases the PR touches now produce the hard break the reference produces, [§6.7](https://spec.commonmark.org/0.31.2/#hard-line-breaks) spec examples 633–647 still pass unchanged, and a few edges (tab + 1 space, just tabs, many tabs + 2 spaces, hard break inside emphasis) all match the reference's call.

Two things worth a look before merge:

**Bug.** The new `trailingHardBreakSpaces` lands between the existing `@type {Resolver}` JSDoc and `resolveAllLineSuffixes`. JSDoc only attaches to the next declaration, so `resolveAllLineSuffixes` ends up un-annotated, and `tsc --strict --checkJs` (which `npm run build` invokes) flags both params as implicit `any`:

```
text.js(184,33): error TS7006: Parameter 'events' implicitly has an 'any' type.
text.js(184,41): error TS7006: Parameter 'context' implicitly has an 'any' type.
```

Move `trailingHardBreakSpaces` above the `@type {Resolver}` block, or below `resolveAllLineSuffixes`, so the annotation lands on its target.

**Suggestion.** The new function repeats the backward walk the existing loop already does. Could fold both into one pass by tracking `trailingSpaces` alongside `size`: count only while a contiguous space run is still in progress, stop on first tab or non-space chunk. Then `tabs ||` becomes `trailingSpaces < 2` and the helper isn't needed. I benched the duplicate walk at +0–3% on a worst-case shape (200 lines, `tab+2sp` every line), so it's style not perf, but the redundancy obscures what's going on.

One small note: [§6.7](https://spec.commonmark.org/0.31.2/#hard-line-breaks) says *"preceded by two or more spaces"* without spelling out what happens when tabs are mixed in (more context in #206). The behavior here matches commonmark.js, which is a defensible interpretation, but a one-line comment above the disqualifier would save the next reader the same dig.

Potentially Out-of-scope FYI: micromark currently consumes the trailing tab into `<br />` (`foo\t  \n` → `<p>foo<br />`), while commonmark.js preserves it as `<p>foo\t<br />`. Not part of this PR, just on the list for later if anyone's chasing exact parity.

wyf027:
Implemented the requested changes in 78664c6:

- Removed the separate `trailingHardBreakSpaces` helper so the `@type {Resolver}` JSDoc attaches to `resolveAllLineSuffixes` again.
- Folded trailing-space tracking into the existing backward walk with a contiguous-space-run marker.
- Added a short comment explaining that only spaces contiguous with the line ending count toward hard break detection.
- Updated the PR body to use the project template.

Local verification:
- `npm run build`
- `npm run format`
- `npm run test-api-dev` (1928 tests)

The remaining canary failures appear to be the same Node 26.2.0 / `yargs` ESM issue also seen on #223/#225, not specific to this PR.

PR #225, opened 2026-05-26
fix: import debug compatibly in development ESM bundle

### LINKED ISSUE #217: micromark: development-only ESM export requires transpiling on Vite dev server
### Initial checklist

- [x] I read the [support docs](https://github.com/micromark/.github/blob/main/support.md)
- [x] I read the [contributing guide](https://github.com/micromark/.github/blob/main/contributing.md)
- [x] I agree to follow the [code of conduct](https://github.com/micromark/.github/blob/main/code-of-conduct.md)
- [x] I searched [issues](https://github.com/search?q=user%3Amicromark&type=issues) and [discussions](https://github.com/orgs/micromark/discussions) and couldn’t find anything (or linked relevant results below)

### Affected package

micromark@4.0.2

### Steps to reproduce

1. Create a Vite app that doesn't optimize any deps by default.
2. Import micromark, or higher-level packages like remark.

### Actual behavior

Observe:

```
create-tokenizer.js:40 Uncaught SyntaxError: The requested module '/node_modules/.pnpm/debug@4.4.3/node_modules/debug/src/browser.js?v=7d7df986' does not provide an export named 'default' (at create-tokenizer.js:40:8)
```

This is because the `development` export condition will be triggered on dev server:

```
  "exports": {
    ".": {
      "development": "./dev/index.js",
      "default": "./index.js"
    },
    "./stream": {
      "development": "./dev/stream.js",
      "default": "./stream.js"
    }
  },
```

(I saw your discussion in https://github.com/vitejs/vite/issues/4815, but it's indeed the current behaviour on Vite 8)

And in development exports, dependency `debug` has its own `browser` export condition which is a CJS module, micromark is a ESM module, the way `create-tokenizer.js` imports `debug` will cause error on Vite.

### Expected behavior

I think preferably the ESM bundle should works in browser without further bundling, like a defensive measure as follows:

```ts
import * as createDebug from 'debug'

const debug = (createDebug.default ?? createDebug)('micromark')
```

Can fix the original issue.

While experimenting, `unified` also had a similar problem with `extend` (cjs), and `hast-util-to-jsx-runtime` with `style-to-js`, both can be patched this way. Not sure about your thoughts on this, I am willing to open a PR if you want.

This will save some work when maintaining Vite configs, because those cjs deps are transitive, consumers have to hardcode them in Vite configs like `["pkg-a > unified", "pkg-a > micromark"]`, especially the dep graph can be complicated. Also considering both CJS deps should have not problems in browser environment, only some workarounds are needed to get it working.

(upstream issue for reference: https://github.com/fuma-nama/fumadocs/issues/3205)

### Runtime

Chrome

### Package manager

pnpm

### Operating system

Darwin Kernel Version 24.6.0

### Build and bundle tools

Vite

wyf027:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This changes the development bundle `debug` import to use a namespace import with a `default` fallback, fixing the Vite dev-server case where `debug` resolves to its CJS `browser` build and does not provide a default ESM export.

This overlaps with #223, which removes the `debug` dependency more broadly. This PR is the narrower compatibility fix for #217; I am fine with closing it if maintainers prefer #223 as the path forward.

Test plan:

* [x] `npm run test-api-dev` (1927 tests pass)

Fixes #217


ChristianMurphy:
This is solving the same problem as https://github.com/micromark/micromark/pull/223

wyf027:
Thanks for pointing out the overlap with #223. I updated this PR body to use the project template and called that out explicitly.

This PR is the narrower compatibility fix for #217, while #223 removes the debug dependency more broadly. I am fine with closing this in favor of #223 if that is the preferred direction.

I also checked the failing CI logs: the failures are from the canary workflow running against Node 26.2.0 and hitting `require is not defined in ES module scope` in `yargs`; the same canary/Node failure pattern is present on #223, so it does not look specific to this PR change.

wyf027:
Closing this in favor of #223. A maintainer noted this PR solves the same problem as #223, and the broader PR remains the active path for that direction. The remaining failing checks here are the Node 26/yargs canary issue already noted above, not a PR-specific fix.

PR #232, opened 2026-09-05
Preserve dotless i in identifier normalization

DePasqualeOrg:
### Initial checklist

* [x] I read the support docs 

### Description of changes

This PR keeps dotless `ı` distinct from ASCII `I`/`i` during identifier normalization, matching Unicode default case folding and preventing reference links from resolving to non-equivalent labels.

It documents the exception and adds regression tests. All 1,928 tests pass.


wooorm:
Not what CM does? https://spec.commonmark.org/dingus/?text=%5BI%5D%3A%20%2Fa%0A%0A%5Bı%5D%3A%20%2Fb%0A%0A%5Ba%5D%5Bi%5D%20%5Bb%5D%5Bı%5D

DePasqualeOrg:
I investigated this using Claude Code, which produced the following analysis based on verified evidence. I’ve also submitted a [PR](https://github.com/commonmark/commonmark.js/pull/308) in commonmark.js to fix this issue there.

---

The dingus runs commonmark.js, and its [`normalizeReference`](https://github.com/commonmark/commonmark.js/blob/2cbde72f5137fcc4b0c829903a13c9500fe3003c/lib/inlines.js#L101-L108) normalizes with `.toLowerCase().toUpperCase()`. That approximates case folding closely enough for `ẞ`/`SS`, but `'ı'.toUpperCase()` is `'I'`, so it collides dotless i with ASCII I. [`micromark-util-normalize-identifier`](https://github.com/micromark/micromark/blob/774a70c6bae6dd94486d3385dbd9a0f14550b709/packages/micromark-util-normalize-identifier/dev/index.js) does the same, which is what this PR changes.

That shortcut came from [commonmark.js#168](https://github.com/commonmark/commonmark.js/issues/168), which replaced a generated case-folding table with it in [`b8fc0c71`](https://github.com/commonmark/commonmark.js/commit/b8fc0c715a7fd4a79f851eac47fa7c43747c998f). The case for it was that the pair “normalize ALL characters in the same way as case fold spec”, measured on Node 12, alongside a size and ASCII performance win. The size and speed arguments hold; the equivalence claim does not, and dotless i is where it fails.

CommonMark specifies [Unicode case fold](https://spec.commonmark.org/0.31.2/#matches) for label matching, and the two are distinct under it: `I` folds to `i`, while `ı` has no fold and stays `ı`.

The two reference implementations disagree on your example:

```markdown
[I]: /a

[ı]: /b

[a][i] [b][ı]
```

cmark, through GitHub, resolves the labels separately:

```html
<p><a href="/a">a</a> <a href="/b">b</a></p>
```

commonmark.js 0.31.2 sends both to `/a`, because the second definition collides with the first and the earlier one wins:

```html
<p><a href="/a">a</a> <a href="/a">b</a></p>
```

The corresponding fix for commonmark.js is submitted as [#308](https://github.com/commonmark/commonmark.js/pull/308).

Dotless i is the only character this affects. Scanning the BMP for characters where `.toLowerCase().toUpperCase()` merges what `casefold()` keeps apart returns `I`/`i` against `ı` and nothing else, once codepoints unassigned in Unicode 16.0 are excluded. So skipping it closes the gap rather than patching one instance of a wider problem.

This PR moves micromark to the behavior the spec describes.

wooorm:
Have you considered the inverse, that there is something to fix in C, or the spec?

DePasqualeOrg:
@wooorm, I'm not deeply familiar with the C implementation or the spec. I encountered this issue while building a markdown editor that requires compatibility between various markdown engines. I've spent over an hour doing due diligence on this issue using Claude Code and Codex, to make sure I'm not wasting anyone's time. Claude Code says this isn't an issue in the C implementation or the spec. Questions like this could be easily resolved if you would use the same tools, informed by your much better familiarity with the subject matter.

wooorm:
You can forward my question to your tools!

DePasqualeOrg:
I did just that, before responding to you.

wooorm:
Hmm, then I don't have an answer yet? "Claude Code says this isn't an issue in the C implementation or the spec." is just regurgitating that they are different. Yes, they are. But it's not an answer whether they should be changed?

DePasqualeOrg:
@wooorm, please stop being difficult. I think it's already clear from what I wrote that I investigated exactly this point. The spec doesn't justify the current behavior in JavaScript.

wooorm:
Maybe there's a reason why the spec author and the authors of `markdown-it` and `marked` and I do things though:

> If we combine those approaches, str.toLowerCase().toUpperCase() is observed to normalize ALL characters in the same way as case fold spec (as tested on node 12, earlier v8 versions have had some issues).

And see https://github.com/commonmark/commonmark.js/issues/168#issuecomment-511806057

DePasqualeOrg:
Again, I think you would benefit a lot from checking this with your own tools. The comment you linked to was already considered while developing this PR. Claude Code's summary is below.

---

That quote is the claim this PR disputes, and the linked comment is the performance argument, which the PR grants.

The claim is testable. Grouping every assigned code point by its `toLowerCase().toUpperCase()` result and asking where default folding (`C`+`F`) disagrees, on node 26, exactly one group in all of Unicode is wrong: `U+0049 U+0069 U+0131` are merged, where folding gives `i` and `ı`. That narrowness is why this is a one-line exception rather than a return to a generated table.

Nor does any profile produce the current behavior: default gives {I, i} · {ı}, Turkic gives {I, ı} · {i}, the shortcut gives {I, i, ı}. Blessing it would mean declaring `ı` equivalent to both `i` and `I`. cmark needs no change either – its table is generated from `CaseFolding.txt`, which has no entry for U+0131.

`markdown-it` and `marked` inherit the same shortcut from the same issue, so they are not independent confirmation.

wooorm:
Fascinating! Do you have any ideas why JavaScript is only different from Unicode case folding for 1 character? That seems... Oddly specific?

DePasqualeOrg:
@jgm's [justification](https://github.com/commonmark/commonmark.js/commit/b8fc0c715a7fd4a79f851eac47fa7c43747c998f) for using `str.toLowerCase().toUpperCase()` was that it "does an accurate unicode case fold", which, as we've seen, is not true.

This is outside my area of expertise, but Claude Code's analysis below is clear on this.

---

The difference is not a coincidence. Lowercasing collapses case downward, and uppercasing then merges characters that share an uppercase but differ in lowercase (`σ`/`ς`, `θ`/`ϑ`). Between them the pair unifies anything related by case – which is what folding does.

So they can only disagree where Unicode deliberately *refuses* to unify a case pair. The default profile does that once: `ı` gets no `CaseFolding.txt` entry, because equating it with `i` is language-specific and belongs to the Turkic tailoring (`0049; T; 0131`). `toUpperCase` is locale-independent and has no such exception, so `'ı'.toUpperCase()` is `'I'`.

That predicts the count. Across every assigned cased code point, U+0131 is the only one with no fold entry whose uppercase moves it out of its own case class.

wooorm:
Really interesting again!
Thanks. I'd like to wait on @jgm and commonmark.js, to see which way the projects should go :)

DePasqualeOrg:
https://github.com/commonmark/commonmark.js/pull/308 has been merged.

